Issue #2, opened 2024-08-09
Import matmul/conv/attention tests from iree/tests/e2e/


ScottTodd:
Could start with what already exists, including the C++ binaries files and CMake build system:

* https://github.com/iree-org/iree/tree/main/tests/e2e/matmul
* https://github.com/iree-org/iree/tree/main/tests/e2e/convolution
* https://github.com/iree-org/iree/tree/main/tools/testing/e2e
* https://github.com/iree-org/iree/blob/main/build_tools/cmake/iree_e2e_generated_runner_test.cmake

Or... run the generator scripts and jump straight to a prototype replacement test runner (pytest? other cmake functions?)

The test generators seem to be using mostly upstream dialects (e.g. `linalg.conv_2d_nchw_fchw`). The build system usage does some fancy things with flags though... should watch out for layering in there. For example:
```cmake
  COMPILER_FLAGS
    "--iree-opt-data-tiling=false"
    "--iree-llvmcpu-enable-scalable-vectorization"
    "--iree-llvmcpu-target-triple=aarch64-unknown-unknown"
    "--iree-preprocessing-pass-pipeline=builtin.module\(util.func\(iree-preprocessing-transpose-matmul-pass{input=lhs}\)\)"
  LABELS
    "requires-arm-sme"
  TARGET_CPU_FEATURES_VARIANTS
    "arm_64:sme:+sve,+sme"
```

See this related thread on Discord: https://discord.com/channels/689900678990135345/1270451599231156266

ScottTodd:
Can also draw on code from https://github.com/nod-ai/rocm-gemm-benchmark/. That has test case generators and checked in .mlir files.

ScottTodd:
Current plan (cc @erman-gurses ):

1. Land https://github.com/iree-org/iree/pull/17751 for interim coverage of attention on CPU (and possibly ROCm)
2. Fork C++ and CMake matmul test code to this repo (including all transitive deps, a new CMake project that depends on the runtime, etc.)
3. Fork C++ and CMake conv and attention test code to this repo
4. Prototype a new test structure (check in generated files, rework CMake/ctest project or use pytest, etc.)

ScottTodd:
Finding lots to clean up while refactoring the matmul tests. Debating starting fresh or cleaning up in-place.

Notes/ideas:

- [ ] https://github.com/iree-org/iree-test-suites/pull/14 (run the generator offline and check in the test case files)
[...]

PR #13, opened 2024-08-16
Begin porting matmul test suite from iree-org/iree.

ScottTodd:
Progress on https://github.com/iree-org/iree-test-suites/issues/2.

This begins to port the tests from https://github.com/iree-org/iree/tree/main/tests/e2e/matmul. The changes here aim to be minimal, but a lot of cleanup is needed to make this sustainable.

* A new CMake project named `iree-test-suites-matmul` is added. We could share this with other test suites that need CMake, or let each test suite create its own. I don't want to overgeneralize yet while so much is still up in the air.
* The CMake project depends on the IREE runtime libraries and header files to build a `iree-e2e-matmul-test` binary. This binary handles preparing inputs, running a compiled program, and comparing outputs against a reference implementation. That may be split into multiple libraries or parts of it may move into Python or another test harness. TBD.
* The CMake project then defines some test cases using `iree_generated_e2e_runner_test`. This may be split into an offline generation step and a test runner other than ctest. TBD.

TODO:

- [x] Fix namespace/symbol collisions in `test_utils.h` (fork further) and include it correctly
- [x] Fix namespace/symbol collisions in `iree_e2e_generated_runner_test.cmake` (fork further)
- [ ] Add a GitHub Actions workflow running the tests
- [x] Structure the CMake project such that it can be used from iree-org/iree in place of the current tests
- [ ] Fork or remove dependencies on the upstream CMake helpers used by `matmul/iree_e2e_generated_runner_test.cmake`?
- [x] Allow pointing at a local IREE, like https://github.com/ScottTodd/iree/tree/infra-e2e-test-suites which deletes the in-tree tests and supporting infra


erman-gurses (approved):
Thanks, Scott, LGTM!

PR #14, opened 2024-08-20
Generate MLIR files for matmul tests.

ScottTodd:
Progress on https://github.com/iree-org/iree-test-suites/issues/2

This runs the generator script using some of the option combinations found in the CMakeLists.txt so we can look at what it generates. This does _not_ set `--transpose_rhs` or `--compilation_info` yet.

Some observations:

* I'm not sure if grouping by "shape" in the source files will continue to be useful. I'd rather split these out into one file per test case. We can add sharding and batching up a level in the test runner. I guess this does let us feed multiple modules in to the compiler at the same time, at least.
* We can decouple compiler flags, test tags, CPU features, runtime flags, and similar properties from these generated test files.
* I would like to also decouple the compilation info parameter - that feels like a meta property that should come through compilation flags, not be part of the test sources.

### DIFF
--- a/matmul/README.md
+++ b/matmul/README.md
@@ -5,11 +5,11 @@ https://github.com/iree-org/iree/tree/main/tests/e2e/matmul.
 
 ## Prerequisites
 
-## Quickstart
-
 First ensure you have the prerequisites from
 https://iree.dev/building-from-source/getting-started/, including CMake, a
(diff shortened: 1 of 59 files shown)


ScottTodd:
Closing in favor of https://github.com/iree-org/iree-test-suites/pull/22

PR #15, opened 2024-08-20
Drop matmul "compilation_info" generator code and test cases.

ScottTodd:
Progress on https://github.com/iree-org/iree-test-suites/issues/2

(This is not yet running on CI anywhere, still refactoring the test suite before passing it back into developer hands)

This was leaking compiler implementation details into test cases. The compiler should be selecting a pipeline on its own based on the target device(s). If a developer wants to try a different pipeline, they can make a code change, introduce a dedicated flag (global/hacky), or use a transform dialect script.

I think this drops some of the rocm test suites, which were only added using explicit compilation_info settings. Leaving adding that back to a follow-up / further refactoring. Editing the 2000 line `CMakeLists.txt` file (generated from a 700 line `BUILD.bazel` file) is error-prone especially with how this file has grown over the years. _If_ we stick to CMake/CTest, it probably makes sense to start the file over from scratch.


kuhar (commented):
In the PR description, could you clarify if we still tests the same matmul shapes (including data types)?

ScottTodd:
> In the PR description, could you clarify if we still tests the same matmul shapes (including data types)?

There's a note there already. This probably drops some coverage. I'm going to continue refactoring the test suite then add coverage back across the full set of backends and test cases in some new format.

ScottTodd:
Ping @erman-gurses ?

erman-gurses (approved):
Based on what we synced before, LGTM.

kuhar (approved):

PR #21, opened 2024-09-04
Rename matmul/ to linalg_ops/, rebase IREE, add test workflow.

ScottTodd:
Progress on https://github.com/iree-org/iree-test-suites/issues/2.

* Renamed top level `matmul/` folder to `linalg_ops/`. We can put `convolution` (linalg) and `attention` (linalg_ext) tests under here too. Maybe other named ops or generics too.
* Rebased on the latest IREE code, including flag renames (though we should drop usage of some flags and prefer to test the main user-facing options)
* Added a workflow that goes through the instructions in `linalg_ops/README.md` and configured both workflows to run on `push` events in addition to the current triggers.


.github/workflows/test_linalg_ops.yml
```diff
@@ -0,0 +1,77 @@
...
+        run: |
+          cmake --build ${BUILD_DIR}
+          cmake --build ${BUILD_DIR} --target iree-test-suites-linalg-ops-deps
+      - name: CTest
+        working-directory: linalg_ops
+        run: |
+          ctest \
+            --test-dir build/ \
+            -R iree-test-suites \
+            --label-exclude "^driver=vulkan$|^driver=metal$|^driver=cuda$|^driver=hip$"
```
saienduri, line 77: Any reason, we don't run on hip too?

saienduri (approved):
Nice

.github/workflows/test_linalg_ops.yml
ScottTodd: Just using a standard GitHub runner with no GPU for this workflow. Latest, we can run these tests here on any runners we want and/or we can fetch the tests from the IREE repo and run there using code from PRs (see https://github.com/iree-org/iree/pull/18291 for a proof of concept)

saienduri: Ok cool, sounds good

PR #33, opened 2024-10-18
Fix mismatch between CMake option names.

ScottTodd:
The `COMPILER_ARGS` option was getting ignored since `COMPILER_FLAGS` is what the helper function was checking.

Probably some copy/paste bug when forking from https://github.com/iree-org/iree/blob/main/build_tools/cmake/iree_e2e_generated_runner_test.cmake where `COMPILER_FLAGS` and `RUNNER_ARGS` are used. I think it makes sense to just call both `ARGS` or `FLAGS`.


erman-gurses:
Thank you, LGTM!

erman-gurses (approved):

PR #35, opened 2024-10-24
Remove xfails from in-tree onnx op test config.

ScottTodd:
This way, we won't need to update the xfails in this repository after nightly releases affect which tests are passing.

Downstream (iree-org/iree) will still use the xfail behavior, while this repository will just report the current status, including all logs for failures.

This also now uploads a HTML report for each run to github actions artifacts. We could host these reports on github pages for easier viewing too.


.github/workflows/test_onnx_ops.yml
```diff
@@ -53,25 +54,31 @@ jobs:
 
       # Run tests and output new config files as needed.
       - name: Run ONNX ops test suite
+        continue-on-error: true
```
ScottTodd, line 57: This workflow will rarely "fail" now. We might want some way to treat _test_ failures as okay, but still report _test runner_ failures (e.g. out of disk space, failed to download files, etc.).

pashu123 (approved):
LGTM!

PR #40, opened 2024-11-19
Update workflows to use latest IREE nightly packages.

ScottTodd:
Package names and the install method changed with https://github.com/iree-org/iree/releases/tag/iree-3.0.0.


ScottTodd:
Phew, no regressions.

PR #44, opened 2024-11-21
Move ONNX vision model tests under model_zoo/validated.

ScottTodd:
Description | Path example
-- | --
Before | `onnx_models/tests/vision/classification_models_test.py`
After | `onnx_models/tests/model_zoo/validated/vision/classification_models_test.py`

This now mirrors the structure in the upstream onnx/models repository more closely: https://github.com/onnx/models/tree/main/validated/vision, in preparation for adding tests for the non-validated models or models from other sources like https://huggingface.co/models.

### DIFF
--- a/onnx_models/README.md
+++ b/onnx_models/README.md
@@ -103,14 +103,16 @@ Each test generates some files as it runs:
 
 ```text
 ├── artifacts
-│   └── vision
+│ └── model_zoo
+│   └── validated
+│     └── vision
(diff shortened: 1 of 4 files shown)


vinayakdsci (approved):
In that case, I guess these imports are okay. But I still think we should shorten them, maybe  in subsequent PRs.

For now, LGTM. Thanks for this patch!

PR #50, opened 2024-11-27
Add a llama 3.1 toy test with cross entropy test

rsuderman:
If we pick the maximum token each decode step we should get a somewhat consistent cross entropy loss. This uses a prebaked llama model. The `irpa` file should stay static but the `mlir` model should update as features are enabled.


sharktank_models/llama3.1/test_llama.py
```diff
@@ -0,0 +1,111 @@
+import iree.compiler
```
ScottTodd, line 1: Please add a license header comment to this file.

sharktank_models/llama3.1/test_llama.py
```diff
@@ -0,0 +1,111 @@
...
+@pytest.fixture
+def llama_mlir():
+    dirname = os.path.dirname(__file__)
+    return f"{dirname}/assets/toy_llama.mlir"
+
+
+@pytest.fixture
+def llama_irpa():
+    dirname = os.path.dirname(__file__)
+    return f"{dirname}/assets/toy_llama.irpa"
```
ScottTodd, line 21: Prefer to use `from pathlib import Path` (https://docs.python.org/3/library/pathlib.html) over `os.path`. This code assumes Unix-style file paths when it doesn't need to.

```suggestion
from pathlib import Path
THIS_DIR = Path(__file__).parent

@pytest.fixture
def llama_mlir():
    return str(THIS_DIR  / "assets" / "toy_llama.mlir")


@pytest.fixture
def llama_irpa():
    return str(THIS_DIR  / "assets" / "toy_llama.irpa")
```

(You can either return a `Path` or a string. I'd prefer to return a `Path` and only convert to string where necessary)

sharktank_models/llama3.1/assets/toy_llama.irpa
ScottTodd, line 1: 3MB IRPA file. Are these "real" weights or splats? Probably real given the file size and size of the model?

We could use Git LFS (https://git-lfs.com/) for .irpa files as we did in https://github.com/nod-ai/SHARK-TestSuite/

https://github.com/nod-ai/SHARK-TestSuite/blob/5a5092f8cf493bb8b6f61e3787eb0d2d49c5071f/iree_tests/.gitattributes#L1-L2

Fine to start without Git LFS at the moment, but worth considering since this will affect the repository size and `git checkout` time.

sharktank_models/llama3.1/test_llama.py
```diff
@@ -0,0 +1,111 @@
...
+        ],
+    )
+
+
+@pytest.fixture
+def compile_llama_gfx1100(llama_mlir):
+    return iree.compiler.compile_file(
+        llama_mlir,
+        extra_args=["--iree-hal-target-device=hip", "--iree-hip-target=gfx1100"],
+    )
```
ScottTodd, line 40: Could load the target from an environment variable if we want to be able to run the tests on different GPUs without needing to edit the source.

sharktank_models/llama3.1/test_llama.py
```diff
@@ -0,0 +1,111 @@
...
+            self.parameters, self.hal, self.binary, config=self.config
+        )
+
+        self.prefill = self.modules[-1].prefill_bs1
+        self.decode = self.modules[-1].decode_bs1
+
+
+@pytest.fixture
+def toy_llama_cpu(compile_llama_cpu, llama_irpa):
+    return ToyLlama(compile_llama_cpu, "local-sync", llama_irpa)
```
ScottTodd, line 67: May want to use `local-task` instead of `local-sync` for the runtime device, depending on how long the test takes to run on CPU and how much multithreading influences the test behavior.

Could also use keyword arguments instead of positional arguments given how arbitrary these variables/strings look:

```suggestion
@pytest.fixture
def toy_llama_cpu(compile_llama_cpu, llama_irpa):
    return ToyLlama(compiled=compile_llama_cpu, device_id="local-task", irpa=llama_irpa)
```

sharktank_models/llama3.1/test_llama.py
```diff
@@ -0,0 +1,111 @@
...
+
+
+@pytest.fixture
+def toy_llama_gfx1100(compile_llama_gfx1100, llama_irpa):
+    return ToyLlama(compile_llama_gfx1100, "hip", llama_irpa)
+
+
+def prefill_cross_entropy(toy_llama, ids):
```
ScottTodd, line 75: Can you add a comment or two to this function? Could be something like

```
# Post-process the results of the prefill function for a more meaningful
# accuracy metric.
# (then maybe some explanation of cross entropy, like a link to some paper or doc)
```

sharktank_models/llama3.1/test_llama.py
```diff
@@ -0,0 +1,111 @@
...
+    ids = ids[0, 1 : ids.shape[1]]
+
+    return torch.nn.functional.cross_entropy(logits, ids, ignore_index=0)
+
+
+def test_prefill_cpu(toy_llama_cpu):
+    ids = numpy.array(
+        [[0, 208, 214, 29, 19, 86, 176, 120, 120, 80, 120, 208, 37, 157, 191, 137]],
+        numpy.int64,
+    )
```
ScottTodd, line 98: Do these numbers map to anything meaningful? Like some text that got processed through a tokenizer?

sharktank_models/llama3.1/test_llama.py
```diff
@@ -0,0 +1,111 @@
...
+    cross_entropy = prefill_cross_entropy(toy_llama_cpu, ids)
+    cross_entropy = cross_entropy.item()
+    assert cross_entropy == pytest.approx(0.589, 1e-3)
```
ScottTodd, line 101: nit: Could add a message along with these asserts

```suggestion
    assert cross_entropy == pytest.approx(0.589, 1e-3), "prefill cross entropy outside of expected tolerance"
```

https://docs.pytest.org/en/stable/how-to/assert.html

sharktank_models/llama3.1/test_llama.py
```diff
@@ -0,0 +1,111 @@
...
+
+
+def test_prefill_gfx1100(toy_llama_gfx1100):
```
ScottTodd, line 104: Can you add a `mark` for test cases that need special hardware? Like `@pytest.mark.plat_rdna3_hip` or `@pytest.mark.target_rocm_gfx1100` or `@pytest.mark.target_rocm`.

See also https://github.com/iree-org/iree/pull/19306#discussion_r1859140473

sharktank_models/llama3.1/test_llama.py
ScottTodd, line 1: General setup notes for this repository:

* Each test suite can have its own workflow file(s) in https://github.com/iree-org/iree-test-suites/tree/main/.github/workflows . This is typically a model for how to run the tests in https://github.com/iree-org/iree, like how [`pkgci_test_onnx.yml` in iree-org/iree](https://github.com/iree-org/iree/blob/41115bba05960e563791ce6ed1af26093f4fab1e/.github/workflows/pkgci_test_onnx.yml#L124-L175) is similar to [`test_onnx_models.yml` in this repo](https://github.com/iree-org/iree-test-suites/blob/main/.github/workflows/test_onnx_models.yml)
* Each test suite should have an overview in the root README: https://github.com/iree-org/iree-test-suites/blob/main/README.md#test-suites
* Each test suite should have its own README file with instructions for how to run the tests (and ideally how to maintain them too)

We don't need all of that in this initial PR, but it would be good to add sooner rather than later

sharktank_models/llama3.1/test_llama.py
```diff
@@ -0,0 +1,111 @@
...
+import torch
+
+page_size = 12288
+block_size = 16
+
+
+@pytest.fixture
+def llama_mlir():
+    dirname = os.path.dirname(__file__)
+    return f"{dirname}/assets/toy_llama.mlir"
```
ScottTodd, line 15: Could add a comment in this file or a README at `sharktank_models/llama3.1` explaining what about this model is a "toy" (number of layers, properties of the weights, etc.) and how to regenerate it

ScottTodd (commented):
Looks good! The top level `sharktank_models` path SGTM. We can shuffle things around later as more tests are added too.

A few mostly structural and style comments.

sharktank_models/llama3.1/test_llama.py
rsuderman: Done.

sharktank_models/llama3.1/test_llama.py
rsuderman: Fixed.

sharktank_models/llama3.1/test_llama.py
rsuderman: Updated.

sharktank_models/llama3.1/test_llama.py
rsuderman: Added

sharktank_models/llama3.1/test_llama.py
rsuderman: Added the workflow and remove the `torch` dep. It brings the e2e testing time to < 30s so it should make the whole process a little more lightweight.

sharktank_models/llama3.1/test_llama.py
rsuderman: Added a readme comment.

sharktank_models/llama3.1/assets/toy_llama.irpa
rsuderman: Updated to use LFS

sharktank_models/llama3.1/test_llama.py
rsuderman: Added environment variable HIP device

sharktank_models/llama3.1/test_llama.py
rsuderman: Marked for `hip` support. We can pull the device from environment variables which makes more sense anyway.

.github/workflows/test_sharktank_models.yml
```diff
@@ -0,0 +1,72 @@
...
+      - name: Run Sharktank models test suite
+        run: |
+          source ${VENV_DIR}/bin/activate
+          pytest sharktank_models/ \
+            -rA \
+            -m "target_cpu" \
+            --timeout=300 \
+            --durations=0 \
+            --html=${HTML_REPORT_PATH} \
+            --self-contained-html
```
ScottTodd, line 66: FYI this contains no log output right now: 
image

Might get some info with `--log-cli-level=info` or the other flags (https://docs.pytest.org/en/stable/how-to/logging.html)

ScottTodd (approved):
Very nice! Thanks for the changes!

sharktank_models/llama3.1/test_llama.py
ScottTodd: So speedy!

.github/workflows/test_sharktank_models.yml
rsuderman: Fixed

sharktank_models/llama3.1/README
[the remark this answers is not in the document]
rsuderman: Done

.github/workflows/test_sharktank_models.yml
ScottTodd: Thanks!
```
------------------------------ Captured log setup ------------------------------
INFO     root:binaries.py:185 Invoke IREE Tool: /home/runner/work/iree-test-suites/iree-test-suites/.venv/lib/python3.11/site-packages/iree/compiler/tools/../_mlir_libs/iree-compile /home/runner/work/iree-test-suites/iree-test-suites/sharktank_models/llama3.1/assets/toy_llama.mlir --iree-input-type=auto --iree-vm-bytecode-module-output-format=flatbuffer-binary --mlir-print-debuginfo --mlir-print-op-on-diagnostic=false --iree-hal-target-device=llvm-cpu --iree-llvmcpu-target-cpu=host
```


image

PR #51, opened 2024-12-02
Add decode test for llama3.1 model

rsuderman:
Include a test for decode that performs cross entropy across the decode steps.

### DIFF
--- a/sharktank_models/llama3.1/test_llama.py
+++ b/sharktank_models/llama3.1/test_llama.py
@@ -39,9 +39,9 @@ def compile_llama_cpu(llama_mlir):
 
 
 @pytest.fixture
-def compile_llama_gfx1100(llama_mlir):
+def compile_llama_hip(llama_mlir):
     if "HIP_TARGET" not in os.environ:
-        return None
+        raise RuntimeError("HIP_TARGET not set")
 
     target_gpu = os.environ["HIP_TARGET"]
     return iree.compiler.compile_file(
@@ -52,9 +52,6 @@ def compile_llama_gfx1100(llama_mlir):
 
 class ToyLlama:
     def __init__(self, compiled, device_id, irpa):
-        if compiled is None:
-            return
-
         paramIndex = iree.runtime.ParameterIndex()
         paramIndex.load(irpa)
         provider = paramIndex.create_provider("model")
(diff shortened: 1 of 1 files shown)


ScottTodd (approved):
Please fix the DCO check: https://github.com/iree-org/iree-test-suites/pull/51/checks?check_run_id=33813712253

Docs: https://iree.dev/developers/general/contributing/#developer-certificate-of-origin

sharktank_models/llama3.1/test_llama.py
```diff
@@ -71,8 +68,49 @@ def __init__(self, compiled, device_id, irpa):
...
+        PADN = N + pad
+
+        ids = numpy.asarray([ids], dtype=numpy.int64)
+        length = numpy.asarray([N], numpy.int64)
+        pages = PADN // block_size
+        pages = numpy.asarray([list(range(pages))], numpy.int64)
+
+        logits = self._prefill(ids[:, :PADN], length, pages, *self._cache)
+        logits = numpy.asarray(logits).astype(numpy.float32)
+        return logits[:, :N]
```
ScottTodd, line 89: It isn't obvious to me at a glance what this line is doing. Would a longer variable name for `N` help? Or a comment here on what the slice is accomplishing?

sharktank_models/llama3.1/test_llama.py
```diff
@@ -71,8 +68,49 @@ def __init__(self, compiled, device_id, irpa):
...
+        self._cache = self.allocate_cache()
+
+        self._prefill = self.modules[-1].prefill_bs1
+        self._decode = self.modules[-1].decode_bs1
+
+    def prefill(self, *, ids):
+        N = len(ids)
+        pad = (16 - N % 16) % 16
+        ids = ids + [0] * pad
+        PADN = N + pad
```
ScottTodd, line 80: Might be worth a comment here (also, why 16? would 32 or another work?)

```suggestion
        # Pad to the next multiple of 16.
        pad = (16 - N % 16) % 16
        ids = ids + [0] * pad
        PADN = N + pad
```

sharktank_models/llama3.1/test_llama.py
```diff
@@ -106,34 +144,48 @@ def cross_entropy(predictions, targets, epsilon=1e-12):
...
-    logits = numpy.asarray(logits).astype(numpy.float32)
-    ids = numpy.asarray(ids)
+def decode_cross_entropy(toy_llama, ids):
+    N = len(ids)
+    logits = toy_llama.decode(ids=ids, start=0)
+    ids = numpy.asarray([ids])
+    logits = logits[0, :-1, :]
+    ids = ids[:, 1:]
+
+    return cross_entropy(logits, ids)
```
ScottTodd, line 163: Out of curiosity, what does `logits` look like before this slicing? What information is discarded?

ScottTodd (approved):

sharktank_models/llama3.1/test_llama.py
rsuderman: Updated the name to make it make more sense. Its just so we pad out to the blocked sequence length.

sharktank_models/llama3.1/test_llama.py
rsuderman: Added a comment about padding

sharktank_models/llama3.1/test_llama.py
rsuderman: Added a comment to clarify. Its just comparing the predicted to the selected token sequence.

PR #52, opened 2024-12-03
Added sharding configurations of tests

rsuderman:
Finished llama 3b tests to include sharding tests. This adds a tp2 sharded version for cpu and hip tests. It also loosens numeric requirements.


sharktank_models/llama3.1/assets/.gitattributes
```diff
@@ -1,2 +1,6 @@
 toy_llama.irpa filter=lfs diff=lfs merge=lfs -text
 toy_llama.mlir filter=lfs diff=lfs merge=lfs -text
+toy_llama_tp2.rank0.irpa filter=lfs diff=lfs merge=lfs -text
+toy_llama_tp2.rank1.irpa filter=lfs diff=lfs merge=lfs -text
+toy_llama_tp2.mlir filter=lfs diff=lfs merge=lfs -text
+toy_llama_tp2.irpa filter=lfs diff=lfs merge=lfs -text
```
ScottTodd, line 6: Want to switch to `*.irpa` and `*.mlir` instead of adding each individual file?

We are also fairly limited (without spending money) in how much we can store with LFS and download

image

Docs: 
* https://docs.github.com/en/billing/managing-billing-for-your-products/managing-billing-for-git-large-file-storage/about-billing-for-git-large-file-storage
* https://docs.github.com/en/billing/managing-billing-for-your-products/managing-billing-for-git-large-file-storage/upgrading-git-large-file-storage


Just something to be careful with >_>

sharktank_models/llama3.1/test_llama.py
```diff
@@ -50,10 +64,39 @@ def compile_llama_hip(llama_mlir):
...
+            "--iree-hal-target-device=hip",
+            "--iree-hal-target-device=hip",
+            f"--iree-hip-target={target_gpu}",
+        ],
+    )
+
+
 class ToyLlama:
-    def __init__(self, compiled, device_id, irpa):
+    def __init__(self, compiled, device_id, irpa, sharding=1):
```
ScottTodd, line 96: Can you add a docstring and maybe type hints for this init function, particularly for the new `sharding` parameter? ("Number of homogeneous devices to shard the model across?")

sharktank_models/llama3.1/test_llama.py
```diff
@@ -22,9 +22,23 @@ def llama_mlir():
...
+    return [str(THIS_DIR / "assets/toy_llama.irpa")]
+
+
+@pytest.fixture
+def llama_tp2_irpa():
+    return [
+        str(THIS_DIR / "assets/toy_llama_tp2.irpa"),
+        str(THIS_DIR / "assets/toy_llama_tp2.rank0.irpa"),
+        str(THIS_DIR / "assets/toy_llama_tp2.rank1.irpa"),
+    ]
```
ScottTodd, line 41: Can you comment a bit on how these sharded .irpa files were generated for the toy model and how they would be similarly generated for the full scale model, e.g. from safetensors or gguf?

Could we reuse the same .irpa file for any number of shards? Do we have tools to split and merge?

If I wanted to regenerate these files, would I run https://github.com/nod-ai/shark-ai/blob/main/sharktank/sharktank/models/llama/toy_llama.py with some `--sharding` option? Or run some tools?

sharktank_models/llama3.1/test_llama.py
```diff
@@ -250,5 +315,75 @@ def test_prefill_decode_hip(toy_llama_hip):
     cross_entropy = prefill_decode_cross_entropy(toy_llama_hip, ids)
     cross_entropy = cross_entropy.item()
     assert cross_entropy == pytest.approx(
-        0.589, 1e-3
+        0.589, 1e-2
+    ), "cross entropy outside of tolerance"
+
+
+@pytest.mark.target_cpu
```
ScottTodd, line 322: You could parameterize these tests across sharding levels (and cpu/hip), rather than copy/paste

https://docs.pytest.org/en/stable/how-to/parametrize.html

rsuderman: I reworked the pytest fixtures to better grow configurations. It keeps the same target / environment variable initialization but makes it easier to expand if we choose to. Should be useful for adding grok variants as well.

sharktank_models/llama3.1/test_llama.py
rsuderman: I added details on generating the files to the `README.md`. Its probably the easiest way to document where / how these are generated.

sharktank_models/llama3.1/assets/.gitattributes
rsuderman: We shouldn't see too much expansion. The biggest issue is that if bandwidth is measured we may encounter issues running on presubmits. 20 MBS of models would be my BS estimate, so we would blow through the limit in 50 PRs which is not that many. Thoughts on how we can consider bandwidth workarounds? I can likely get them a bit smaller but not likely that much tinier.

sharktank_models/llama3.1/test_llama.py
rsuderman: done.

sharktank_models/llama3.1/assets/.gitattributes
ScottTodd: The costs aren't _that_ high, it's just annoying to need to think about it.

50GB of bandwidth per month costs $5/month, $60 per year. That's pennies compared to what we spend on other cloud infra. We'd just need to configure billing in iree-org.

cc @amd-chrissosa 

---

Alternatives are:

* Use regular git instead of git LFS
* Use huggingface
* Use Azure
* Use AWS
* Use GCS

(I'd prefer to use something that community members can upload to, so git LFS or huggingface and not a cloud project)

amd-chrissosa: Drive by comment, preference against Git. Git is awful for blob storage and penalizes all checkouts even if the blob isn't being used.  However, we should probably have a standard answer here for blob storage across all projects. IMO Azure, AWS, GCS could be used if we can tie contributor access to credentials. We don't have a good story for that atm though.

ScottTodd: Yeah, hence trying to use Git LFS... but GitHub sets the free tier laughably low (1GB of storage, 1GB of bandwidth / month) and the pricing model is pay-as-you-go using "data packs"... very strange.

sharktank_models/llama3.1/test_llama.py
```diff
@@ -109,20 +95,56 @@ def decode(self, *, ids, start=0):
...
 
-@pytest.fixture
-def toy_llama_hip(compile_llama_hip, llama_irpa):
-    return ToyLlama(compiled=compile_llama_hip, device_id="hip", irpa=llama_irpa)
+def hip_flags(sharding):
+    if "HIP_TARGET" not in os.environ:
+        raise RuntimeError("HIP_TARGET not set")
+
+    target_gpu = os.environ["HIP_TARGET"]
+    return ["--iree-hal-target-device=hip"] * sharding + [
```
sogartar, line 120: We would be running all shards on the same device. May not catch some problems where we mistakenly mix buffers on different devices where we don't expect DMA.
At some point may be worth running the test on a multi-device system.

sogartar (approved):

sharktank_models/llama3.1/README.md
```diff
@@ -5,3 +5,29 @@ motivation is to have an executable version of the model to validate both
 compilation and numerics of execution. This is generated using using sharktank
 with the default coefficients. Future updates for the `mlir` file should
 maintain the same `irpa` values while only updating the executed code.
+
+This files are generated by the following
+
+```
```
ScottTodd, line 11: nit: add syntax highlighting

```suggestion
```bash
```

sharktank_models/llama3.1/assets/.gitattributes
ScottTodd: Bandwidth usage is at half the quota already x_x
image

Let's figure out the billing side offline, @amd-chrissosa ?

ScottTodd (approved):

sharktank_models/llama3.1/test_llama.py
rsuderman: It should be good enough for base testing.

PR #55, opened 2024-12-10
Rework hip and multi device configuration

rsuderman:
Changed multi device configuration to support separate device compilation instead of multiples of the same device. Add compile flag to avoid numerical error known on `hip`.

### DIFF
--- a/sharktank_models/llama3.1/test_llama.py
+++ b/sharktank_models/llama3.1/test_llama.py
@@ -121,6 +121,7 @@ def hip_flags(sharding):
     target_gpu = os.environ["HIP_TARGET"]
     return [f"--iree-hal-target-device=hip[{i}]" for i in range(sharding)] + [
         f"--iree-hip-target={target_gpu}",
+        # TODO: Remove once #19347 is addressed
         "--iree-codegen-block-dynamic-dimensions-of-contractions=false",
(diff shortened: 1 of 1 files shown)


kumardeepakamd (approved):
looks good

sharktank_models/llama3.1/test_llama.py
```diff
@@ -117,8 +119,9 @@ def hip_flags(sharding):
         raise RuntimeError("HIP_TARGET not set")
 
     target_gpu = os.environ["HIP_TARGET"]
-    return ["--iree-hal-target-device=hip"] * sharding + [
-        f"--iree-hip-target={target_gpu}"
+    return [f"--iree-hal-target-device=hip[{i}]" for i in range(sharding)] + [
+        f"--iree-hip-target={target_gpu}",
+        "--iree-codegen-block-dynamic-dimensions-of-contractions=false",
```
ScottTodd, line 124: > Add compile flag to avoid numerical error known on `hip`.

Can you link to some context for this flag in a comment? Maybe a link to an issue where discussion happened? If this flag becomes the default or it is no longer needed, I'd like future developers to know when the flag is safe to remove here.

ScottTodd (approved):

PR #68, opened 2025-01-15
[NFC] Fix a typo in README.md (HTMl -> HTML)

hanhanW:
It turns `l` from lowercase to uppercase.

### DIFF
--- a/sharktank_models/README.md
+++ b/sharktank_models/README.md
@@ -89,7 +89,7 @@ built as part of the [shark-ai project](https://github.com/nod-ai/shark-ai).
     pytest -n 4
     ```
 
-* Create an HTMl report using https://pytest-html.readthedocs.io/en/latest/index.html
+* Create an HTML report using https://pytest-html.readthedocs.io/en/latest/index.html
 
     ```bash
     pytest --html=report.html --self-contained-html --log-cli-level=info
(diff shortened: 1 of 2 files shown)


sharktank_models/README.md
```diff
@@ -89,7 +89,7 @@ built as part of the [shark-ai project](https://github.com/nod-ai/shark-ai).
     pytest -n 4
     ```
 
-* Create an HTMl report using https://pytest-html.readthedocs.io/en/latest/index.html
+* Create an HTML report using https://pytest-html.readthedocs.io/en/latest/index.html
```
ScottTodd, line 92: Can you fix in this other file too? https://github.com/iree-org/iree-test-suites/blob/c47d13c0730a51beed3bef6128e7a61a80f85ce9/onnx_models/README.md?plain=1#L91

I also think it may be time soon to merge some of the docs and utility scripts for the test suites using pytest. I wanted to keep them standalone at the start, but it would be useful to be able to run multiple types of tests in a single invocation and have consistent options between all of them.

ScottTodd (approved):

sharktank_models/README.md
hanhanW: Oh, I did not notice that because this is created when I'm reading the README.md. I verified that there are no `HTMl` in the repo with the fix. Done!

ScottTodd: Yeah, there's a fair bit of copy/paste at the moment, typos and all 😅

PR #71, opened 2025-01-17
Implement caching for onnx/models git LFS files.

ScottTodd:
Progress on https://github.com/iree-org/iree-test-suites/issues/6.

This adds a caching layer which allows developers and persistent CI runners to avoid needing to redownload source `.onnx` files.

## Details

* The cache location defaults to `${IREE_TEST_FILES}/iree-test-suites` if `IREE_TEST_FILES` is set, or `~/.cache/iree-test-suites/` otherwise. This can be overridden with the custom `--cache-dir=/path/to/cache` pytest option. Several of our persistent CI machines set the `IREE_TEST_FILES` environment variable already.
* The cache is implemented as a local git clone of the https://github.com/onnx/models repository, which uses [Git Large File Storage (LFS)](https://git-lfs.com/) to store large files. When a file is requested by a test, the cache layer runs `git lfs pull` in the local clone to fetch the latest version of the file and then it creates a symlink from the cache directory to the test working directory. This usage should be pretty similar to what huggingface_hub provides: https://huggingface.co/docs/huggingface_hub/guides/manage-cache.

## Testing

Tested in iree-org/iree on persistent runners here:

* Cold cache: https://github.com/iree-org/iree/actions/runs/12838451019/job/35804050925#step:8:22

    ```
    ---------------------------- live log sessionstart -----------------------------
    INFO     onnx_models.conftest:conftest.py:96 Using cache directory: '/home/esaimana/iree_tests_cache/iree-test-suites'
    INFO     onnx_models.cache:cache.py:115 Setting up GitHub repository 'onnx/models'
    INFO     onnx_models.cache:cache.py:117 Checking for working 'git lfs' (https://git-lfs.com/)
    INFO     onnx_models.cache:cache.py:136 Cloning https://github.com/onnx/models.git into '/home/esaimana/iree_tests_cache/iree-test-suites/onnx_models'
    Cloning into '/home/esaimana/iree_tests_cache/iree-test-suites/onnx_models'...
    ```

* Warm cache: https://github.com/iree-org/iree/actions/runs/12838451019/job/35804127583#step:8:22

    ```
    ---------------------------- live log sessionstart -----------------------------
    INFO     onnx_models.conftest:conftest.py:96 Using cache directory: '/home/esaimana/iree_tests_cache/iree-test-suites'
    INFO     onnx_models.cache:cache.py:115 Setting up GitHub repository 'onnx/models'
    INFO     onnx_models.cache:cache.py:117 Checking for working 'git lfs' (https://git-lfs.com/)
    INFO     onnx_models.cache:cache.py:122 Directory '/home/esaimana/iree_tests_cache/iree-test-suites/onnx_models' already exists
    ```

(The rest of the logs are currently the same)

### DIFF
--- a/onnx_models/cache.py
+++ b/onnx_models/cache.py
@@ -114,8 +114,9 @@ class GitHubLFSRepositoryCacheScope(CacheScope):
     def setup_github_repository(self, repository_name: str, clone_method: str):
         logger.info(f"Setting up GitHub repository '{repository_name}'")
 
-        logger.info("Checking for working 'git lfs' (https://git-lfs.com/)")
-        subprocess.run("git lfs env", capture_output=True, check=True)
+        # Command does not exist on GitHub-hosted Linux runners?
+        # logger.info("Checking for working 'git lfs' (https://git-lfs.com/)")
+        # subprocess.run("git lfs env", capture_output=True, check=True)
 
         # Skip if the directory already exists (and is a git directory).
         if self.local_repository_dir.is_dir():
--- a/onnx_models/cache.py
+++ b/onnx_models/cache.py
@@ -114,15 +114,14 @@ class GitHubLFSRepositoryCacheScope(CacheScope):
     def setup_github_repository(self, repository_name: str, clone_method: str):
         logger.info(f"Setting up GitHub repository '{repository_name}'")
 
-        # Command does not exist on GitHub-hosted Linux runners?
-        # logger.info("Checking for working 'git lfs' (https://git-lfs.com/)")
-        # subprocess.run("git lfs env", capture_output=True, check=True)
+        logger.info("Checking for working 'git lfs' (https://git-lfs.com/)")
+        subprocess.run(["git", "lfs", "env"], capture_output=True, check=True)
 
         # Skip if the directory already exists (and is a git directory).
         if self.local_repository_dir.is_dir():
             logger.info(f"Directory '{self.local_repository_dir}' already exists")
             subprocess.run(
-                "git rev-parse --is-inside-work-tree",
+                ["git", "rev-parse", "--is-inside-work-tree"],
                 cwd=self.local_repository_dir,
                 capture_output=True,
                 check=True,
@@ -136,7 +135,7 @@ class GitHubLFSRepositoryCacheScope(CacheScope):
             remote_url = f"git@github.com:{repository_name}.git"
         logger.info(f"Cloning {remote_url} into '{self.local_repository_dir}'")
         subprocess.run(
-            f"git clone {remote_url} {self.local_repository_dir}", check=True
+            ["git", "clone", remote_url, self.local_repository_dir], check=True
         )
 
     def pull_lfs_file(self, file_relative_path: str):
--- a/onnx_models/cache.py
+++ b/onnx_models/cache.py
@@ -20,7 +20,19 @@ class CacheScope(abc.ABC):
 
     @abc.abstractmethod
     def get_file(self, relative_path: str) -> Path:
-        """Get the path to a file loaded from the cache."""
+        """Gets the local path to a file, populating the cache as needed.
+
+        Implementations of this class are responsible for performing any setup,
+        download, or other tasks needed to get a local file system path for the
+        file.
+        """
+
+    # TODO(scotttodd): more [abstract] methods?
+    # def print_stats(self):
+    # def prune_cache(self, max_size_in_bytes: int):
+    # def clear_cache(self, max_size_in_bytes: int):
+    # def warm_cache_for_file(self, relative_path: str):
+    # def get_files(self, relative_paths: list[str]) -> list[Path]:
 
 
 class CacheManager:
@@ -62,7 +74,7 @@ class CacheManager:
         self.cache_scopes[scope.scope_name] = scope
 
     def get_file_in_cache(self, scope_name: str, relative_path: str) -> Path:
-        """Gets the path to a file in the cache, if it exists."""
+        """Gets the path to a file in the cache, downloading as needed, if it exists."""
         logger.info(f"Getting file from {scope_name} cache: {relative_path}")
         if scope_name not in self.cache_scopes:
             raise ValueError(f"Unknown cache scope: '{scope_name}'")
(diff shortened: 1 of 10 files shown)


onnx_models/cache.py
```diff
@@ -0,0 +1,167 @@
...
+
+    Each registered scope can independently manage a set of cached files.
+    Symlinks from the cache are created into a common working directory. For example:
+
+    cache_dir_1/
+      onnx_models/    <--- GitHubLFSRepositoryCacheScope for https://github.com/onnx/models
+        validated/vision/classification/
+          mnist/
+            model/
+              mnist-12.onnx
```
ScottTodd, line 27: One thing at a time, but I'll probably want to actually fetch the `.tar.gz` files like https://github.com/onnx/models/blob/main/validated/vision/classification/mnist/model/mnist-12.tar.gz and then unpack them, since those archives include inputs and outputs already generated and validated for us. Could just have the archives cached and then unpack at test time.... orrrr let the cache be responsible for that too. See what huggingface_hub has for extra files ("assets") here: https://huggingface.co/docs/huggingface_hub/guides/manage-cache#caching-assets.

There will be file name conflicts if I extract all of those files into the same flat directory though, in either the cache or working directory 🤔. May want to ditch the current "subdirectory" concept and instead create one artifacts directory per test case.

onnx_models/cache.py
```diff
@@ -0,0 +1,167 @@
...
+            os.remove(working_subdirectory_file)
+        elif working_subdirectory_file.exists():
+            logger.warning("  Non-symlink file exists. Replacing with a symlink")
+            os.remove(working_subdirectory_file)
+        os.symlink(src=file_in_cache, dst=working_subdirectory_file)
+        return working_subdirectory_file
+
+
+class CacheScope(abc.ABC):
+    """Abstract base class for a cache scope."""
```
zjgarvey, line 85: It might be useful to put in the docstring more info about what a child class needs to implement. It's not clear from reading this class definition that a child class needs to implement not only the retrieval of warm cache file paths, but also the downloading/extracting of such files. Perhaps this can be resolved by adding something like:

```python
@abc.abstractmethod
def setup_file(self, relative_path: str):
    """Download or generate the cache file if not present"""
```

However, I'm not sure how to enforce compatibility between this function and the output of `get_file`. Maybe it would just be better to keep this base class simple and add a bit more documentation to indicate that `get_file` should also handle initializing the cache file when not present.

onnx_models/cache.py
zjgarvey: Brainstorming for this:

Maybe you could have the `CacheManager` handle this by pluralizing to `get_files_in_working_directory`, which could return a dict with keys `['model', 'metadata', 'inputs', 'golden_outputs']` in case those are available in the cache.

zjgarvey (commented):
I like the direction this is heading. Factoring out cache management is something e2eshark could benefit from too, since `OnnxModelInfo` <https://github.com/nod-ai/SHARK-TestSuite/blob/4797eae93979854df5ef0bf8cefeb4f2f6c215b6/alt_e2eshark/e2e_testing/framework.py#L45> currently handles too much in one class.

All of my comments are brainstorming/nits. I hope some of them are helpful. Let me know if you want me to re-review. I'm going to leave this review as a comment for now to discuss some of the ideas and hear your feedback.

onnx_models/conftest.py
```diff
@@ -72,13 +71,40 @@ def pytest_addoption(parser):
...
+    logger.info(f"Using cache directory: '{cache_dir}'")
+    cache_dir.mkdir(parents=True, exist_ok=True)
+    # Cache context for https://github.com/onnx/models.
+    cache_manager = CacheManager(working_directory=ARTIFACTS_ROOT)
+    onnx_models_cache = GitHubLFSRepositoryCacheScope(
+        scope_name="model_zoo",
+        cache_dir=cache_dir,
+        repository_name="onnx/models",
+    )
+    cache_manager.cache_scopes.append(onnx_models_cache)
```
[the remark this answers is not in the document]
ScottTodd: I was thinking along similar lines too. Adding a function like that `append_scope` (or `register_scope`) encapsulates the member variable, which would let it change the data structure from a list to a dict as it wants.

onnx_models/cache.py
ScottTodd: Sounds good, will write more docs.

> It's not clear from reading this class definition that a child class needs to implement not only the retrieval of warm cache file paths, but also the downloading/extracting of such files.

The interface is that a cache scope returns a path on a local disk for a given [remote] path. As part of that, the scope may need to download a file, authenticate with a server, or perform some other operations.

onnx_models/conftest.py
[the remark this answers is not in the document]
ScottTodd: Added `register_scope` function.

onnx_models/cache.py
```diff
@@ -0,0 +1,167 @@
...
+    def __init__(self, working_directory: Path):
+        self.cache_scopes = []
+        self.working_directory = working_directory
+
+    def get_file_in_cache(self, scope_name: str, relative_path: str) -> Path:
+        """Gets the path to a file in the cache, if it exists"""
+        logger.info(f"Getting file from {scope_name} cache: {relative_path}")
+        for cache_scope in self.cache_scopes:
+            if cache_scope.scope_name == scope_name:
+                return cache_scope.get_file(relative_path)
```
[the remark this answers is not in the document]
ScottTodd: Switched to Dict.

onnx_models/cache.py
ScottTodd: Added some more docs. How's this look now?

onnx_models/cache.py
ScottTodd: Leaning towards one directory per test case, with the test runner (`conftest.py`) or some onnx util class handling extracting from `.tar.gz` so the cache only needs to download files. Teaching the cache how to extract files and then remove the original archives _could_ save on disk space used, but I'd rather keep the implementation simple right now.

Simple cache, duplicate storage:
```
cache_dir
  test_case_1_archive.tar.gz (git lfs understands how to version this)

working_dir
  test_case_1
    archive.tar.gz (symlink)
    model.onnx (from archive)
    input.pb (from archive)
```

Minimal storage:
```
cache_dir
  [deleted after extraction] test_case_1_archive.tar.gz
  test_case_1
    model.onnx (from archive)
    input.pb (from archive)

working_dir
  test_case_1
    model.onnx (symlink)
    input.pb (symlink)
```

With the minimal storage approach, the cache would need to understand that the extracted files don't need to be redownloaded. That isn't _really_ an issue with the onnx models, or even HF models, since the files are pretty stable and we don't need the absolute latest versions all the time, but I like how the default approach _does_ automatically get latest versions with no extra bookkeeping on our end.

ScottTodd:
> All of my comments are brainstorming/nits. I hope some of them are helpful. Let me know if you want me to re-review. I'm going to leave this review as a comment for now to discuss some of the ideas and hear your feedback.

Thanks! Addressed some of the surface level comments. Want to re-review? I think I'm just about ready to get these tests running on presubmit in https://github.com/iree-org/iree now.

I have more ideas/goals about the "test configs", XFAILs, and other ways that developers and users interface with test suites.I would like to be reporting current status in an easy to understand way. This repository and SHARK-TestSuite should be feeding into release notes and public project status trackers so users can see which programs are known to be working at which versions. We have plenty of the building blocks for that here and in https://github.com/nod-ai/e2eshark-reports/ already.

onnx_models/cache.py
zjgarvey: Yeah, the simple cache approach seems best to me (and seems to likely have minimal storage in cache when not running the tests, assuming the compressed file is actually reasonably compressed).

zjgarvey (approved):
This looks good to me! Thanks for looking through and addressing the comments so quickly.

PR #75, opened 2025-01-31
Add test for sharktank CLIP text model

sogartar:


rsuderman:
You should make sure these binary files are added use `git-lfs`. They looks like they are being committed to the repo itself.

PR #77, opened 2025-02-14
Initial migration of iree regression tests

geomin12:
Migrating [threshold tests](https://github.com/iree-org/iree/tree/main/experimental/regression_suite/shark-test-suite-models) and [benchmark tests](https://github.com/iree-org/iree/tree/main/experimental/benchmarks/sdxl) to this repository and also making improvements

Improvements:
- Made a generalized python script for both threshold tests and benchmark tests, so for current and future models/submodels, you just need to add/update a JSON file instead of duplicating code
- Resolving [issue 19984](https://github.com/iree-org/iree/issues/19984) by specifying file path [here](https://github.com/geomin12/iree-test-suites/blob/da44d58a316018c2a87da182609ca4819e2cd50a/sharktank_models/test_suite/regression_tests/sdxl/punet_int8_fp8.json#L69) and correctly getting the right path [here](https://github.com/geomin12/iree-test-suites/blob/da44d58a316018c2a87da182609ca4819e2cd50a/sharktank_models/test_suite/regression_tests/test_model_threshold.py#L93)
- Improving / creating local development script to run these tests

Things left to do in future PR(s):
- Creating github action script that runs everything (in a midnight run or PR checks) and runs specific selected models through an action workflow_dispatch
- Integrating Hugging Face model retrieval, adding Google's benchmark script and onboarding flux, according to this [PR](https://github.com/nod-ai/shark-ai/pull/870#discussion_r1956435014)


.github/workflows/test_pkgci_regression_run.yml
geomin12, line 1: this will be cleaned up and updated in a future PR

sharktank_models/test_suite/benchmarks/run_benchmarks_for_model.py
geomin12, line 1: this may be converted to a bash script, in future PR

sharktank_models/test_suite/build_tools/setup_venv.py
ScottTodd, line 1: Shouldn't need this file. Delete?

The setup_venv script is used in the IREE repository to install dev packages built by the pkgci.yml workflow and stored in github workflow artifacts. In this repository, we would either install from nightly releases, install from pypi, or clone the IREE repository and use its copy of the setup_venv.py script for dev packages. We shouldn't need to maintain a forked version of the script here.

Installing packages here should be more like
```
pip install --find-links https://iree.dev/pip-release-links.html --pre iree-base-compiler iree-base-runtime
```

Or that can be put in a requirements file and then installed, like here:
https://github.com/iree-org/iree-test-suites/blob/f8f2771c932322a2a813a51a9e6ef71ca7eb1f80/.github/workflows/test_sharktank_models.yml#L45-L55 https://github.com/iree-org/iree-test-suites/blob/f8f2771c932322a2a813a51a9e6ef71ca7eb1f80/sharktank_models/requirements-iree.txt#L1-L9

sharktank_models/test_suite/ireers_tools/artifacts.py
```diff
@@ -0,0 +1,234 @@
...
+            t.total = tsize
+        t.update((b - last_b[0]) * bsize)
+        last_b[0] = b
+
+    return update_to
+
+
+@functools.cache
+def get_artifact_root_dir() -> Path:
+    root_path = os.getenv("IREE_TEST_FILES", default=str(Path.cwd())) + "/artifacts"
```
ScottTodd, line 42: Similar comment about `Path.cwd()`. Could root on `sharktank_models/test_suite/` (get path to the current script file then walk up the tree as needed)

sharktank_models/test_suite/ireers_tools/artifacts.py
```diff
@@ -0,0 +1,234 @@
+# Copyright 2023 The IREE Authors
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+from typing import Any, Callable, Collection, Dict, Union
+import functools
+from pathlib import Path
+from tqdm import tqdm
```
ScottTodd, line 10: This looks unused, could remove then add back if it does get used somewhere.

https://tqdm.github.io/ (it wraps iterables to low a progress meter)

sharktank_models/test_suite/ireers_tools/fixtures.py
```diff
@@ -0,0 +1,83 @@
...
+            "-o",
+            str(vmfb_path),
+            str(source.path),
+        ]
+        + IREE_COMPILE_QOL_FLAGS
+        + flags
+    )
+    print("Exec:", " ".join(exec_args))
+    start_time = time.time()
+    subprocess.run(exec_args, check=True, capture_output=True, cwd=vmfb_path.parent)
```
ScottTodd, line 53: Duplicate code with `run_iree_command` in `sharktank_models/test_suite/benchmarks/run_benchmarks_for_model.py`? We have lots of versions of this in the repository now, would prefer to only fork again if truly needed.

Same for `iree_run_module` and `iree_benchmark_module` in this file.

sharktank_models/test_suite/regression_tests/README.md
```diff
@@ -0,0 +1,52 @@
...
+    -rpfE \
+    --log-cli-level=info \
+    --capture=no \
+    --timeout=600 \
+    --durations=0 \
+    --model-name sd3 \
+    --submodel-name vae
+```
+
+### Required and optional fields for the JSON model file
```
ScottTodd, line 28: Docs here look duplicated with the benchmarks? Should there be one doc shared between each, instead of duplicating?

sharktank_models/test_suite/regression_tests/sd3/clip.json
```diff
@@ -0,0 +1,70 @@
...
+            "source": "https://sharkpublic.blob.core.windows.net/sharkpublic/sai/sd3-prompt-encoder/inference_output.0.bin",
+            "value": "2x154x4096xf32"
+        },
+        {
+            "source": "https://sharkpublic.blob.core.windows.net/sharkpublic/sai/sd3-prompt-encoder/inference_output.1.bin",
+            "value": "2x2048xf32"
+        }
+    ],
+    "real_weights": "https://sharkpublic.blob.core.windows.net/sharkpublic/sai/sd3-prompt-encoder/real_weights.irpa",
+    "mlir": "https://sharkpublic.blob.core.windows.net/sharkpublic/sai/sd3-prompt-encoder/model.mlir",
```
ScottTodd, line 39: Let's aim for each model to have instructions for regenerating it, like https://github.com/iree-org/iree-test-suites/tree/main/sharktank_models/llama3.1#example-llama-model-with-toy-implementation

sharktank_models/test_suite/regression_tests/sd3/clip.json
```diff
@@ -0,0 +1,70 @@
...
+        "--iree-opt-outer-dim-concat=true",
+        "--iree-hip-waves-per-eu=2",
+        "--iree-llvmgpu-enable-prefetch",
+        "--iree-dispatch-creation-enable-aggressive-fusion",
+        "--iree-dispatch-creation-enable-fuse-horizontal-contractions=true",
+        "--iree-opt-aggressively-propagate-transposes=true",
+        "--iree-codegen-llvmgpu-use-vector-distribution=true",
+        "--iree-execution-model=async-external",
+        "--iree-preprocessing-pass-pipeline=builtin.module(iree-preprocessing-transpose-convolution-pipeline,iree-preprocessing-pad-to-intrinsics{pad-target-type=conv})"
+    ],
```
ScottTodd, line 61: Similar to another comment thread, could have one test file per test configuration, rather than `cpu_compiler_flags` and `rocm_compiler_flags`. We have multiple in-tree backends / test configurations and downstream users have out-of-tree backends too. Ideally the tests should be runnable using an out of tree configuration, without needing to edit the test suite itself. That means that configuration files should just have generic names like "compiler flags", "run flags", and "threshold args" and the test suite itself shouldn't contain special logic for any particular backend or configuration.

sharktank_models/test_suite/regression_tests/test_model_threshold.py
```diff
@@ -0,0 +1,196 @@
...
+
+
+class TestModelThreshold:
+    @pytest.fixture(autouse = True)
+    @classmethod
+    def setup_class(self, pytestconfig):
+        self.model_name = pytestconfig.getoption("model_name")
+        self.submodel_name = pytestconfig.getoption("submodel_name")
+
+        file_name = f"{Path.cwd()}/sharktank_models/test_suite/regression_tests/{self.model_name}/{self.submodel_name}.json"
```
ScottTodd, line 57: another brittle use of `Path.cwd()`, recommend using the path to the current script file and then walking up/down from there

sharktank_models/test_suite/setup.py
```diff
@@ -0,0 +1,31 @@
...
+        "numpy",
+        "pytest",
+        "pytest-xdist",
+        "pytest-depends",
+        "pytest-retry",
+        "pytest-timeout",
+        "pytest-xdist",
+        "pytest-check",
+        "tabulate",
+        "tqdm",
```
ScottTodd, line 27: Unused, remove?

sharktank_models/test_suite/setup.py
```diff
@@ -0,0 +1,31 @@
+# Copyright 2023 The IREE Authors
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+from setuptools import find_namespace_packages, setup
+
+setup(
+    name=f"iree-regression-suite",
```
ScottTodd, line 10: FYI I've set up other test suites in this repository to not use a `setup.py`, since that requires an extra step to run the tests.

Without a package (which this setup.py defines), importing uses relative paths like this:
https://github.com/iree-org/iree-test-suites/blob/f8f2771c932322a2a813a51a9e6ef71ca7eb1f80/onnx_models/tests/model_zoo/validated/vision/body_analysis_models_test.py#L11

That makes files harder to author, but it does save some setup when running the tests.

I have been thinking about extracting enough test suite utilities into a common package like this `iree-regression-suite` that would sit at the top level of https://github.com/iree-org/iree-test-suites. This could be a good time for that (follow-up work, not in this PR)

ScottTodd (changes requested):
A mix of low level and high level comments. I think we can land and iterate for most of these, but I do want to keep a high bar in test suite code and not overspecialize on rocm testing. Maybe not all in this PR, but soon.

sharktank_models/test_suite/setup.py
geomin12: yeah i agree! i thought `setup.py` was the preferred format, but I like the requirements.txt route a lot more, and i will add those dependencies to the requirements.txt!

ScottTodd: setup.py creates a project (https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#configuring-your-project), which lets you pip install and then reuse that project from outside of that directory. That's useful for common tools with stable APIs shared between multiple folders or repositories.

Most of the test suites so far have been collections of utility scripts that _could_ be structured into a project but didn't yet reach that level of maturity. I think creating a project is a good idea.... IF we roll it out across multiple test suites and scope it appropriately for that. Having just one of the test suites create a project in a subfolder doesn't quite meet that bar IMO.

I can try to find some other frameworks etc. in the ecosystem with test suite projects for more ideas...

ScottTodd: This could be another good reference: https://github.com/EricSchles/drifter_ml

sharktank_models/test_suite/regression_tests/sd3/clip.json
ScottTodd: Another thing to watch for is that tests should run with default flags and possibly also advanced flag configurations like these. That keeps us honest about the out of the box experience being functional and these flags not being load bearing for correctness. For benchmarks, it makes more sense to measure using flag configurations... within reason.

sharktank_models/test_suite/regression_tests/README.md
ScottTodd: Okay, keeping two similar READMEs sounds good to me in that case.

sharktank_models/test_suite/benchmarks/README.md
```diff
@@ -0,0 +1,46 @@
...
+| benchmark_min_warmup_time        | required | float   | The minimum warm up time for the benchmark test                                                                              |
+| golden_time_tolerance_multiplier | optional | object  | An object of tolerance multipliers, where the key is the sku and the value is the multiplier, (ex: `{"mi250": 1.3}`)         |
+| golden_time                      | optional | object  | An object of golden times, where the key is the sku and the value is the golden time in ms, (ex: `{"mi250": 100}`)           |
+| golden_dispatch                  | optional | object  | An object of golden dispatches, where the key is the sku and the value is the golden dispatch count, (ex: `{"mi250": 1602}`) |
+| golden_size                      | optional | object  | An object of golden sizes, where the key is the sku and the value is the golden size in bytes, (ex: `{"mi250": 2000000}`)    |
+| specific_rocm_chip_to_ignore     | optional | array   | An array of chip values, where the benchmark tests will ignore the chips specified                                           |
```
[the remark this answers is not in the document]
geomin12: my assumption was that we would want to test on all chips and in rare cases, decide to exclude specific chips for certain tests

although with the other target-specific options, would folks find more value if they could specify which options / chips to allow?

sharktank_models/test_suite/setup.py
geomin12: Sounds good, I think once the test_suite is all merged into one (or even the entire `iree-test-suite`, there would be value in the `setup.py`, but for this PR, i'll update requirements.txt and remove this

and then next PR, i'll get it all combined and set that up

sharktank_models/test_suite/build_tools/external_test_files/attention_and_matmul_spec_punet_mi300.mlir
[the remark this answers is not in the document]
ScottTodd: I'd read through https://github.com/iree-org/iree/issues/19214 and check the current status with @kuhar .

> We will provide default tuning specs for key architectures like gfx942. Each architecture will have its own tuning spec file. The initial target is performant SDXL on gfx942.

So this use case seems like exactly what that is designed for.

sharktank_models/test_suite/regression_tests/sd3/clip.json
geomin12: I had no idea (beginner thoughts) that there were more options than CPU + ROCM! I agree, I may start working on making it more dynamic for unit tests so that it doesn't have to rely a backend (esp with threshold tests, it specifically tests CPU then ROCM)

For benchmark tests, those are becoming a bit more agnostic than the threshold tests

sharktank_models/test_suite/build_tools/external_test_files/attention_and_matmul_spec_punet_mi300.mlir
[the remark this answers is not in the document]
geomin12: oh beauty! let me reach out

sharktank_models/test_suite/regression_tests/sd3/clip.json
ScottTodd: Docs: https://iree.dev/guides/deployment-configurations/

Our group at AMD has been focused on ROCm/HIP on datacenter GPUs, but we still invest in other backends and have users (including downstream) using Vulkan, CUDA, Metal, CPU, custom backends, etc. Ideally much of our infrastructure is applicable across all backends (including out of tree ones) on a wide range of models, while still providing coverage for the specialized configurations and "halo models" that are the bulk of our team's direct investments.

geomin12: ah this is making sense 👁️👁️, okay cool let me work on making it a bit more than just CPU + ROCM

sharktank_models/test_suite/benchmarks/run_benchmarks.py
```diff
@@ -0,0 +1,37 @@
...
+# Licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+import subprocess
+import  os
+from pathlib import Path
+import argparse
+
+parser = argparse.ArgumentParser()
```
ScottTodd, line 12: Same comment about `__main__` and about a wrapper script being a code smell - the underlying pytest should be runnable on its own _or_ the wrapper script should provide enough value on top that users will want to start with it. For example, a wrapper script could enumerate available devices on the current system and then run _just those benchmarks/tests_. If I try to run benchmarks on a system with an AMD w7900 GPU instead of a MI300 GPU, I should get w7900 benchmarks, not mi300 benchmarks - and same for an NVIDIA 1080/2080/3080/etc.

sharktank_models/test_suite/regression_tests/run_threshold_test.py
ScottTodd, line 1: Ah. This is getting discovered by pytest as a test file since its file name ends with `_test.py`, so the existing github actions workflow is trying to run it and failing: https://github.com/iree-org/iree-test-suites/actions/runs/13421189316/job/37493971926?pr=77#step:6:22. Adding a `__main__` check might help, or you might want to move utility scripts into a subdirectory or change their names so pytest doesn't treat them as tests.

https://docs.pytest.org/en/stable/example/pythoncollection.html

sharktank_models/test_suite/benchmarks/test_model_benchmark.py
```diff
@@ -138,60 +110,79 @@ def setup_class(self, pytestconfig):
...
+            # ROCM or CPU specific configurations
+            self.compile_flags = data.get("compile_flags", [])
+            self.benchmark_additional_flags = []
+            if type_of_chip == "rocm":
+                self.file_suffix = f"{type_of_chip}_{rocm_chip}"
+                self.compile_flags += [
+                    "--iree-hal-target-backends=rocm",
+                    f"--iree-hip-target={rocm_chip}"
+                ]
+                self.benchmark_additional_flags += ["--device_allocator=caching"]
```
ScottTodd, line 136: (for follow-up refactoring)
I think benchmarks should have to opt-in for this flag, instead of having the test runner unconditionally add it.

sharktank_models/test_suite/benchmarks/test_model_benchmark.py
```diff
@@ -138,60 +110,79 @@ def setup_class(self, pytestconfig):
...
+                self.device = "hip"
+                
+            elif type_of_chip == "cpu":
+                self.file_suffix = "cpu"
```
ScottTodd, line 140: Another code smell from the overspecialization on rocm - "chip" means something very specific for rocm. Other backends don't have that concept or use other words.

Instead prefer to have each test configuration bring its own lists of flags and supporting files, as in, for example https://github.com/iree-org/iree-test-suites/tree/main/onnx_models/configs. Then this code could instead read the lists from the configuration and forward them through to the tests/benchmarks - no logic needed in the test framework itself.

ScottTodd (commented):
Thanks for the changes. This is looking fairly close to ready for merging I think, but I don't want to lose track of the refactoring from hardcoding support for only "rocm" and also "cpu" to a generic "configuration", though that can be handled in follow-up changes.

We do need the existing github actions workflow to continue passing with this change though.

sharktank_models/test_suite/regression_tests/README.md
```diff
@@ -6,24 +6,22 @@
...
-    --durations=0 \
-    --model-name sd3 \
-    --submodel-name vae
-```
+| Argument Name | Default value | Description                                                                                                                                      |
+| ------------- | ------------- | ------------------------------------------------------------------------------------------------------------------------------------------------ |
+| --model       | sdxl          | Runs threshold tests for a specific model                                                                                                        |
+| --submodel    | \*            | If specified, the threshold tests will run for a specific submodel (ex: `--submodel clip`). If not specified, it will run tests on all submodels |
+| --sku         | mi300         | The threshold tests will run on this sku and retrieve golden values from the specified sku                                                       |
+| --rocm-chip   | gfx942        | The threshold tests will run on this ROCM chip                                                                                                   |
```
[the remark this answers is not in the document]
geomin12: this will be done in the next PR! making it similar to onnx_ops tests and making it more agnostic (like benchmark tests)

sharktank_models/test_suite/benchmarks/run_benchmarks.py
geomin12: ah! i see, this makes more sense! yeah, i think i structured these tests to be too dependent on the ROCM (my bad, i was learning from the old code)

but yes, in following PRs, i'll restructure it to be dependent on the system! although i think this generalization work is a great start towards that idea

ScottTodd: No worries, similar code has been sprinkled all over the place. Doing this sort of refactoring helps bring that back into focus. Thanks for taking the time to make things better!

sharktank_models/test_suite/regression_tests/sd3/clip.json
geomin12: will be done in next PR, particularly with making this accessible and usable by folks!

ScottTodd (approved):

Issue #26, opened 2024-09-12
Make it easier to share workflow steps with other repositories


ScottTodd:
Forking this issue from https://github.com/nod-ai/SHARK-TestSuite/issues/288.

This workflow https://github.com/iree-org/iree-test-suites/blob/main/.github/workflows/test_onnx_ops.yml has some duplicated code at https://github.com/iree-org/iree/blob/main/.github/workflows/pkgci_test_onnx.yml. 

Most of the duplication is boilerplate and keeping it synced across commits is a bit of a chore. See https://docs.github.com/en/actions/using-workflows/avoiding-duplication for the options available to us.

I'm imagining something downstream like

```yml
jobs:
  test_onnx:
    steps:
      - name: Check out external TestSuite repository
        uses: actions/checkout@v4.1.7
        with:
          repository: nod-ai/SHARK-TestSuite
          ref: v1
          path: SHARK-TestSuite
          submodules: false
          lfs: false
      - name: Run onnx tests
         uses: SHARK-TestSuite/.github/workflows/test_onnx.yml
         with:
           pytest_flags: ...
           config_file: ...
```

or 
```yml
test_onnx:
  uses: iree-org/test-suite@v1  # or nod-ai/iree-test-suite
  with:
    pytest_flags: ...
    config_file: ...    
```

where the workflow would include running tests, checking for diffs in the config file and uploading the new file(s), etc.

---

### New thoughts since filing that:

For local and CI usage, an entry point script (could be `pytest`) that runs _all test suites_ here would be helpful.

Some things that change from run to run or machine to machine:
* Available and/or selected hardware and APIs to use (GPUs connected / visible / requseted, use Vulkan, use CUDA, use ROCm, use CPU, etc.
* Test expectations (which should be passing vs which are known failing, are deviations from that okay or should they fail the run?)
* How to shard, parallelize, log, etc.

PR #79, opened 2025-02-21
Adding composite github action and generalization work

geomin12:
Updates:
- Adding composite action, so `iree` pkgci regression test yml can just call this file and run tests
- Adding `setup.py` for project since clip, llama, quality tests and benchmarks share dependencies
- Some reorganization
- Generalization work for quality tests and benchmark tests

Next todos:
- Make the update in iree to use this action
- for now, i have followed the same format for displaying results in README, i'm going to look into the database and sending data to that. from what i've gathered, it seems like the dashboard was in progress but got de-prioritized so not sure how much data will be visible 

Progress on https://github.com/iree-org/iree-test-suites/issues/26


geomin12:
@ScottTodd one question! In terms of documentation for generating models (like llama or clip), do they pull those `mlir` files from hugging face? i only see `onnx` files! so not sure on how to generate their own models and using them in this test repo in terms of specifics

Of course, I can be quite broad and just mention bring your MLIR or already compiled VMFB and you can configure it with benchmark tests, but not sure if folks want more specifics

sharktank_models/README.md
```diff
@@ -29,15 +29,9 @@ built as part of the [shark-ai project](https://github.com/nod-ai/shark-ai).
...
     source .venv/bin/activate
-    python -m pip install -r requirements.txt
+    python -m pip install -e .
     ```
 
-    * To use IREE from nightly pre-release Python packages:
-
-        ```bash
-        python -m pip install -r requirements-iree.txt
-        ```
```
ScottTodd, line 39: I think we should keep this and the "use a custom version" option on equal footing.

That will involve keeping the `requirements-iree.txt` file around and pulling these deps out of the new `setup.py`:

```diff
    install_requires=[
        "azure-storage-blob",
-       "iree-base-compiler",
-       "iree-base-runtime",
```

If someone is working with a source build they will have some packages and tools already configured in their environment and having the test package force an install from PyPI may overwrite what they have.

sharktank_models/requirements-iree.txt
```diff
@@ -1,9 +0,0 @@
-# Requirements for using IREE from nightly packages.
-
-# Include base requirements.
--r requirements.txt
-
---find-links https://iree.dev/pip-release-links.html
---pre
-iree-base-compiler
-iree-base-runtime
```
ScottTodd, line 9: The first two lines here are important. Moving these package requirements into `setup.py` without these qualifiers will result in only stable releases being installed, when test suites and presubmit tests need the flexibility to install dev or nightly packages.

https://iree.dev/reference/bindings/python/#installing-iree-packages

See how the workflow now installs stable packages:

```diff
      - name: Install IREE nightly release Python packages
        run: |
          source ${VENV_DIR}/bin/activate
-         python3 -m pip install -r sharktank_models/requirements-iree.txt
+         python3 -m pip install -e sharktank_models/
```

https://github.com/iree-org/iree-test-suites/actions/runs/13467282112/job/37635538303?pr=79#step:5:26
```
 Collecting iree-base-compiler (from sharktank_model_tests==0.1.dev1)
  Downloading iree_base_compiler-3.2.0-cp311-cp311-manylinux_2_27_x86_64.manylinux_2_28_x86_64.whl.metadata (1.3 kB)
```

When it should be installing nightly packages: https://github.com/iree-org/iree-test-suites/actions/runs/13459516010/job/37611203923#step:5:47

```
Collecting iree-base-compiler (from -r sharktank_models/requirements-iree.txt (line 8))
  Downloading https://github.com/iree-org/iree/releases/download/iree-3.3.0rc20250221/iree_base_compiler-3.3.0rc20250221-cp311-cp311-manylinux_2_27_x86_64.manylinux_2_28_x86_64.whl (72.0 MB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 72.0/72.0 MB 15.1 MB/s eta 0:00:00
```

sharktank_models/setup.py
```diff
@@ -0,0 +1,33 @@
...
+        "pytest",
+        "pytest-check",
+        "pytest-html",
+        "pytest-reportlog",
+        "pytest-retry",
+        "pytest-timeout",
+        "pytest-xdist",
+        "tabulate",
+    ],
+    extras_require={},
```
ScottTodd, line 32: If you wanted, you could put an extras for the IREE requirements like https://github.com/iree-org/iree/blob/3b7d9dd587a36919489b106c8d1ad41cca11a356/compiler/setup.py#L519-L524 then if you installed with `sharktank_model_tests[iree]` instead of just `sharktank_model_tests` it would fetch the IREE packages too.

I don't think `--find-links` and `--pre` can be passed for individual requirements that way though.

.pre-commit-config.yaml
ScottTodd, line 1: This could be moved to its own PR. Prefer to keep changes focused on just one thing at a time to make review easier and faster.

sharktank_models/action.yml
```diff
@@ -0,0 +1,59 @@
+# Copyright 2025 The IREE Authors
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+name: IREE Test Suite Regression Test
```
ScottTodd, line 7: Probably want "sharktank models" in this name too. It might have also been easier to start with one of the other test suites since I think we'll want all of them to provide composite actions too.

On that note, can you tag https://github.com/iree-org/iree-test-suites/issues/26 in the PR description?

Like: 

```
Progress on https://github.com/iree-org/iree-test-suites/issues/26
```

sharktank_models/action.yml
```diff
@@ -0,0 +1,59 @@
...
+inputs:
+  model:
+    description: "The model to run threshold and benchmark tests"
+    required: true
+  sku:
+    description: "Type of SKU to test"
+    required: true
+  backend:
+    description: "Type of backend to use"
+    required: true
```
ScottTodd, line 18: Only one model at a time? We could pass configuration files as input to generalize this, like how the ONNX tests are customized:

* https://github.com/iree-org/iree/blob/3b7d9dd587a36919489b106c8d1ad41cca11a356/.github/workflows/pkgci_test_onnx.yml#L107
* https://github.com/iree-org/iree/blob/3b7d9dd587a36919489b106c8d1ad41cca11a356/.github/workflows/pkgci_test_onnx.yml#L26-L66
* https://github.com/iree-org/iree/tree/main/tests/external/iree-test-suites/onnx_ops

(That would also requiring decoupling the SKU/backend/etc. from the core test suite code too, which is worth doing since it would enable out of tree backends to use these tests too)

sharktank_models/action.yml
```diff
@@ -0,0 +1,59 @@
...
+  run_benchmarks:
+    description: "If flag is set to true, benchmark tests will run"
+    type: boolean
+    required: true
+
+runs:
+  using: "composite"
+  steps:
+    - name: Checkout the code
+      uses: actions/checkout@v2
```
ScottTodd, line 28: Update this to closer to latest and also pin action versions to commit hashes rather than use tags. Tags can (and often are) changed outside our control, weakening our supply chain

https://github.com/iree-org/iree-test-suites/blob/43ffb5e37c3f72f60d91e498f82419c6a559825b/.github/workflows/test_sharktank_models.yml#L40-L41

More info: https://github.com/ossf/scorecard/blob/main/docs/checks.md#pinned-dependencies

sharktank_models/action.yml
```diff
@@ -0,0 +1,59 @@
...
+
+    - name: Install Python dependencies
+      shell: bash
```
ScottTodd, line 31: You can set the default shell to bash for the entire action

sharktank_models/action.yml
```diff
@@ -0,0 +1,59 @@
...
+    
+    - name: Run compilation and threshold tests
+      shell: bash
+      run: |
+        python ${GITHUB_ACTION_PATH}/sharktank_models/regression_tests/run_thresholds.py \
+          --model=${{ inputs.model }} \
+          --sku=${{ inputs.sku }} \
+          --backend=${{ inputs.backend }}
+    
+      # Using bash conditionals because if statements in composite actions are not yet supported: https://github.com/actions/runner/blob/main/docs/adrs/0549-composite-run-steps.md#if-condition
```
ScottTodd, line 42: 🤯 how is such a basic feature not supported?? Aaaaah.

.github/workflows/test_sharktank_models.yml
```diff
@@ -66,7 +66,8 @@ jobs:
             --log-cli-level=info \
             --html=${HTML_REPORT_PATH} \
             --self-contained-html \
-            --ignore=sharktank_models/test_suite
+            --ignore=sharktank_models/benchmarks \
+            --ignore=sharktank_models/regression_tests
```
ScottTodd, line 70: The split here where `.github/workflows/test_sharktank_models.yml` and `sharktank_models/action.yml` run different sets of tests is likely to be confusing. What will it take to merge both into a single test suite?

Setting up a composite action for the other test suites and proving that out across repositories may be easier since those test suites don't have such open questions at the moment.

sharktank_models/README.md
ScottTodd, line 1: > @ScottTodd one question! In terms of documentation for generating models (like llama or clip), do they pull those `mlir` files from hugging face? i only see `onnx` files! so not sure on how to generate their own models and using them in this test repo in terms of specifics. like do they bring their MLIR files and write some python tests like `clip` or `llama`?
> 
> Of course, I can be quite broad and just mention bring your MLIR or already compiled VMFB and you can configure it with benchmark tests, but not sure if folks want more specifics

Each model should have a README explaining how to generate the files, like https://github.com/iree-org/iree-test-suites/tree/main/sharktank_models/llama3.1

sharktank is the model development toolkit in shark-ai. It contains model implementations and connections between repositories like huggingface and the IREE compiler. You can use it to get from, for example, https://huggingface.co/meta-llama/Llama-3.1-8B to .mlir and .irpa files that the IREE compiler and runtime can use.

sharktank_models/README.md
```diff
@@ -97,3 +91,13 @@ built as part of the [shark-ai project](https://github.com/nod-ai/shark-ai).
...
     See also
     https://docs.pytest.org/en/latest/how-to/output.html#creating-junitxml-format-files
+
+## Running threshold tests
+
+Please refer to [this file](regression_tests/README.md) to run threshold tests
+
+## Running benchmark tests
+
+Please refer to [this file](benchmarks/README.md) to run benchmark tests
```
ScottTodd, line 101: nit: when writing documentation, prefer to put a short, descriptive phrase as the link text, not phrases like "this document": https://developers.google.com/style/link-text. Here I would insert the document titles like `Please refer to [Regression tests](...)`

also... "threshold tests" doesn't carry much meaning to me. I would typically expect "correctness tests" "quality tests" "accuracy tests" or a similar term here in contrast to "benchmark tests"

sharktank_models/benchmarks/run_benchmarks.py
```diff
@@ -0,0 +1,68 @@
...
+        "pytest",
+        THIS_DIR / "test_model_benchmark.py",
+        "--log-cli-level=info",
+        "--timeout=600",
+        "--retries=0",
+    ]
+    
+    with open("job_summary.md", "a") as job_summary, open("job_summary.json", "w+") as content:
+        print(f"{sku.upper()} {model.upper()} Complete Benchmark Summary:\n", file=job_summary)
+        json.dump({}, content)
```
ScottTodd, line 44: May want to add this file to `.gitignore` if it isn't already so if a developer runs this locally it doesn't pollute their git index.

Another thing you can do is detect if you are running in a github action and change the behavior based on that. Plenty of ways to do that. Stella added one recently in https://github.com/nod-ai/TheRock/blob/main/build_tools/teatime.py with an environment variable that the script checks for.

ScottTodd (changes requested):
Nice, happy to see composite actions getting added.

A common thread in my review comments here is keeping changes small and focused. It can be tempting to bundle changes together but that makes reviewing and addressing feedback take more effort on both sides.

sharktank_models/action.yml
geomin12: i know! i was surprised that it wasn't available yet

sharktank_models/action.yml
```diff
@@ -0,0 +1,59 @@
...
+      uses: actions/checkout@v2
+
+    - name: Install Python dependencies
+      shell: bash
+      run: pip install -e ${GITHUB_ACTION_PATH}/sharktank_models
+    
+    - name: Run compilation and threshold tests
+      shell: bash
+      run: |
+        python ${GITHUB_ACTION_PATH}/sharktank_models/regression_tests/run_thresholds.py \
```
ScottTodd, line 37: Probably also want to (optionally?) give this composite action a Python venv to work within, allowing callers of the action to install the IREE packages however they want ahead of the action running. We also use some persistent self-hosted runners that should be installing into virtual environments instead of using the system-provided Python

Maybe add a `VENV_DIR` argument that is the root path of a virtual environment to use, then each step in the composite action that uses Python can run `source ${VENV_DIR}/bin/activate` as here: https://github.com/iree-org/iree/blob/54e1770334768c54bd35132618b90d6acdbfde3c/.github/workflows/pkgci_test_onnx.yml#L95-L112

We've also tried some other ways to make "activating a virtual environment" persistent across job steps but I haven't seen one that is robust enough. Could be worth another search.

sharktank_models/action.yml
geomin12: lame but composite actions doesn't support this yet: https://github.com/orgs/community/discussions/18597

sharktank_models/setup.py
geomin12: Yeah, ideally it would be neat if the setup.py could do this, but after some brief searching, seems like we'll have to do one setup.py for the base, then the requirements-iree.txt on top of that

sharktank_models/README.md
geomin12: Ah okay, this helps clarify a lot! Thanks

.github/workflows/test_sharktank_models.yml
geomin12: Yeah good point, I think they can be merged together when the "complete generalization" of quality + benchmark test are done. 

Once the first composite action is setup and in action, I will work on transitioning these tests over to their own separate composite actions since it seems like the `test_clip` and `test_llama` are testing different things compared to the quality + benchmark tests.

sharktank_models/action.yml
geomin12: The ONNX code is fantastic, didn't know you could use the `pytest.Item` parameter inside of conftest, although this does require more generalization work that is expected for the next PR (more generalizing quality tests and benchmark tests)

For this PR, I'll just get the composite action working in iree along with the first generalization, then continue working on generalization work!

.github/workflows/test_sharktank_models.yml
ScottTodd: I'd prefer for them to share a single action / pytest call if at all possible. We're already too fragmented across workflows in other repositories and this sort of infra should be helping unify.

For example, these are all doing very similar things:

* https://github.com/nod-ai/shark-ai/blob/main/.github/workflows/ci-llama-quick-tests.yaml
* https://github.com/nod-ai/shark-ai/blob/main/.github/workflows/ci-sdxl.yaml
* https://github.com/nod-ai/shark-ai/blob/main/.github/workflows/ci-sglang-benchmark.yml
* https://github.com/nod-ai/shark-ai/blob/main/.github/workflows/ci_eval.yaml

We should have tests for 100s of models and we'll never get there if we have one bespoke workflow per model.

geomin12: i'll get cooking on that 🧑‍🍳, i can probably use the onnx format to at least cover the json generation, and pytest will pick that up!

sharktank_models/README.md
```diff
@@ -97,3 +97,35 @@ built as part of the [shark-ai project](https://github.com/nod-ai/shark-ai).
...
+
+In order to generate and compile MLIR files to compile, run quality tests and benchmarking tests, please run the following the following commands:
+
+This example generates IRPA and MLIR files for Llama, please look in [Shark AI Models](https://github.com/nod-ai/shark-ai/tree/main/sharktank/sharktank/models) to see which models you can generate
+
+```
+git clone https://github.com/nod-ai/shark-ai.git
+
+cd shark-ai/sharktank
+python3 -m pip install .
```
ScottTodd, line 121: I see that these instructions are forked from `llama3.1/README.md`. They are fine as-is, but we should generally steer users and test suites towards package builds instead of source builds. That would mean replacing

```
git clone https://github.com/nod-ai/shark-ai.git
cd shark-ai/sharktank
python3 -m pip install .
```

with

```
python3 -m pip install sharktank
```

for stable releases or this for nightly releases, following https://github.com/nod-ai/shark-ai/blob/main/docs/nightly_releases.md:

```
python3 -m pip install sharktank -f https://github.com/nod-ai/shark-ai/releases/expanded_assets/dev-wheels --pre
```

sharktank_models/action.yml
ScottTodd: (Unresolving this conversation)

This should be the responsibility of the caller, so dev or nightly packages can be used depending on what the caller wants:

```yml
- name: Setting up virtual environment and install Python dependencies
      shell: bash
      run: |
        python -m venv ${GITHUB_WORKSPACE}/venv 
        source ${GITHUB_WORKSPACE}/venv/bin/activate
        pip install -e ${GITHUB_ACTION_PATH}/sharktank_models
        pip install -r ${GITHUB_ACTION_PATH}/sharktank_models/requirements-iree.txt
```

To fix:

1. Add a new input to the workflow named something like `VENV_DIR`
2. In each calling workflow, create the venv using a path like `${GITHUB_WORKSPACE}/.venv` and install IREE requirements
3. In this workflow, install `pip install -e ${GITHUB_ACTION_PATH}/sharktank_models`

I _think_ you could actually keep this code with only minor edits if you want the default behavior to be installing requirements that this action provides, since the pip commands should no-op if used on an already existing directory or a venv that already has packages installed. However, I would actually prefer to be explicit and require the caller to do that setup so that this workflow doesn't unexpectedly fall back to something that the caller doesn't expect.

sharktank_models/benchmarks/README.md
```diff
@@ -19,7 +19,7 @@ Argument options for the script
 | --model       | sdxl          | Runs benchmark tests for a specific model                                                                                                        |
 | --filename    | \*            | If specified, the benchmark tests will run for a specific filename (ex: `--filename clip`). If not specified, it will run tests on all filenames |
 | --sku         | mi300         | The benchmark tests will run on this sku and retrieve golden values from the specified sku                                                       |
-| --rocm-chip   | gfx942        | The benchmark tests will run on this ROCM chip                                                                                                   |
+| --backend   | gfx942        | The benchmark tests will run on this backend                                                                                                   |
```
ScottTodd, line 22: gfx942 isn't a backend, it's a (amdgpu) target chip. Why the flag name change here?

sharktank_models/benchmarks/run_benchmarks.py
```diff
@@ -0,0 +1,68 @@
...
+import sys
+import tabulate
+import json
+
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--model", type=str, default="sdxl")
+    parser.add_argument("--filename", type=str, default="*")
+    parser.add_argument("--sku", type=str, default="mi300")
+    parser.add_argument("--backend", type=str, default="gfx942")
```
ScottTodd, line 20: Same here - gfx942 isn't a backend.

* These are compiler backends: https://iree.dev/guides/deployment-configurations/#compiler-target-backends
* These are runtime drivers/devices: https://iree.dev/guides/deployment-configurations/#runtime-hal-driversdevices
* These are target chips for the rocm backend: https://iree.dev/guides/deployment-configurations/gpu-rocm/#compile-a-program

sharktank_models/benchmarks/README.md
geomin12: ah! my bad, i thought chip / backend could be interchangeable since chip is only applicable to amdgpus, but seems like chip is the correct definition.

sharktank_models/benchmarks/run_benchmarks.py
geomin12: okay fantastic, those links help and clarify that!

geomin12:
@ScottTodd sorry for the insane amount of file changes, most are just reorganization and decoupling

sharktank_models/action.yml
```diff
@@ -0,0 +1,92 @@
...
+        source ${{ inputs.VENV_DIR }}/bin/activate
+        pip install -e ${GITHUB_ACTION_PATH}/sharktank_models
+        pip install -r ${GITHUB_ACTION_PATH}/sharktank_models/requirements-iree.txt
+    
+    - name: Run sharktank specific model tests
+      shell: bash
+      if: contains(inputs.MATRIX_NAME, "model")
+      run: |
+        source ${{ inputs.VENV_DIR }}/bin/activate
+        HIP_TARGET=${{ inputs.GPU }} pytest iree-test-suites/sharktank_models/ \
```
ScottTodd, line 52: Looks like `HIP_TARGET` here could be set in the environment the same way as below, instead of as a prefix to the pytest command.

```yml
      env:
        ROCM_CHIP: ${{ inputs.CHIP }}
        SKU: ${{ inputs.SKU }}
```

sharktank_models/quality_tests/README.md
```diff
@@ -0,0 +1,39 @@
...
+| outputs                             | optional | array   | An array of objects that provides the output blob and the expected output value (ex: `{"source" :"", "value": ""}`, the value field is optional) |
+| real_weights                        | optional | string  | URL that provides the real weights blob                                                                                                          |
+| mlir                                | required | string  | URL that provides the MLIR blob                                                                                                                  |
+| pipeline_mlir                       | optional | string  | URL that provides the MLIR blob                                                                                                                  |
+| device                              | optional | string  | The device to run the threshold tests on                                                                                                                  |
```
ScottTodd, line 28: nit: looks like the final `|` characters in this table are not aligned. You could omit them or add the right number of spaces

image

sharktank_models/quality_tests/sd3/clip_cpu.json
```diff
@@ -0,0 +1,54 @@
+{
+    "inputs": [
+        {
+            "source": "https://sharkpublic.blob.core.windows.net/sharkpublic/sai/sd3-prompt-encoder/inference_input.0.bin",
+            "value": "1x77x2xi64"
+        },
```
ScottTodd, line 6: We could move these inputs and outputs into Git LFS instead of Azure 🤔. The `real_weights.irpa` files should probably stay on Azure or come from HuggingFace though, since they'll be 100MB+

Also note that for outputs, we can't generally expect bit exactness or even floating point similarity across backends. For some of these integration tests we will want a quality metric like the cross entropy comparison in https://github.com/iree-org/iree-test-suites/blob/main/sharktank_models/llama3.1/test_llama.py or the perplexity comparison in https://github.com/nod-ai/shark-ai/blob/main/sharktank/tests/evaluate/perplexity_iree_test.py.

ScottTodd (approved):
Looks good as an incremental improvement. Thanks for also filing https://github.com/iree-org/iree-test-suites/issues/81

sharktank_models/quality_tests/sd3/clip_cpu.json
ScottTodd: (no action needed in this PR though)

ScottTodd:
Looks good once the DCO check is resolved.

Since you are a member of iree-org, I suggest setting up signing with an SSH key by following the instructions at https://iree.dev/developers/general/contributing/#crypographically-signing-commits. As a bonus, all of your commits across GitHub will then include the "verified" marker :)

geomin12:
> Looks good once the DCO check is resolved.
> 
> Since you are a member of iree-org, I suggest setting up signing with an SSH key by following the instructions at https://iree.dev/developers/general/contributing/#crypographically-signing-commits. As a bonus, all of your commits across GitHub will then include the "verified" marker :)

yes i was just thinking that. i keep forgetting!!

PR #82, opened 2025-02-27
Adding pre-commit and one cleanup item

geomin12:
Quick PR to add pre-commit, pre-commit cleanup and making inputs for sharktank composite action optional (as some fields will be used, not all)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,118 @@
+# Pre-commit (https://pre-commit.com) configuration for assorted lint checks.
+#
+# See https://pre-commit.com/hooks.html for more hooks.
+
+exclude: "third_party/"
+
+repos:
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v3.2.0
+    hooks:
+      - id: check-merge-conflict
+
+      - id: check-yaml
+        # * Extensions can't be included in the mkdocs schema, so skip checking
+        #   https://github.com/squidfunk/mkdocs-material/issues/6378
+        # * clang-format files use `---` to split for multiple languages,
+        #   resulting in errors like `expected a single document in the stream`
+        exclude: "mkdocs.yml|.clang-format"
+
+      - id: end-of-file-fixer
+        exclude_types: ["image", "jupyter"]
+        exclude: ".*\\.excalidraw"
+
+      - id: trailing-whitespace
+        exclude_types: ["image", "jupyter"]
+
+  - repo: https://github.com/psf/black
+    rev: 23.3.0
+    hooks:
+      - id: black
+        name: Run Black to format Python files
+
+  - repo: https://github.com/pre-commit/mirrors-clang-format
+    # Loosely track the most recent versions in
+    #   * Runner images: https://github.com/actions/runner-images/
+    #   * Editor extensions: https://github.com/microsoft/vscode-cpptools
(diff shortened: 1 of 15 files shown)


litert_models/requirements.txt
```diff
@@ -11,4 +11,4 @@ pytest-xdist
 # ai-edge-litert
 # tflite-runtime
 kagglehub
-tensorflow
+tensorflow==2.18.0
```
geomin12, line 14: i don't like the explicit version calling, but according to the [pip --pre flag](https://pip.pypa.io/en/stable/cli/pip_install/#cmdoption-pre), it should be installing the stable release

but for some reason, it is installing the [newest prerelease tensorflow](https://pypi.org/project/tensorflow/#history)
`  Downloading tensorflow-2.19.0rc0-cp311-cp311-manylinux_2_17_x86_64.manylinux2014_x86_64.whl.metadata (4.1 kB)
`

and this is causing test failures! so this current release should cause it to pass

ScottTodd: Can you move this change to a separate PR? Keep PRs focused on one thing at a time.

The failures here are pre-existing: https://github.com/iree-org/iree-test-suites/actions/workflows/test_litert_models.yml?query=branch%3Amain. I have an explanation and some recommended next steps over at https://github.com/iree-org/iree/issues/19777#issuecomment-2679096000. Limiting the version here is fine too, but we should follow what the docs (which I also wrote recently) recommend: https://iree.dev/guides/ml-frameworks/tflite/#prerequisites

geomin12: sounds good! yeah, i'll make sure to keep PRs scoped

.pre-commit-config.yaml
```diff
@@ -0,0 +1,24 @@
...
+        name: Run Black to format Python files
+
+  - repo: https://github.com/igorshubovych/markdownlint-cli
+    rev: v0.41.0
+    hooks:
+      - id: markdownlint
+        name: Run markdownlint on .md files
+        args: ["--config", "docs/.markdownlint.yml"]
+        files: "docs/website/.*.md"
+        exclude: "mlir-dialects/!(index).md"
```
ScottTodd, line 24: Can remove this hook or edit the paths. There is no `docs/` folder in this repo.

ScottTodd (approved):

.pre-commit-config.yaml
```diff
@@ -0,0 +1,15 @@
...
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v3.2.0
+    hooks:
+      - id: check-merge-conflict
+
+  - repo: https://github.com/psf/black
+    rev: 23.3.0
+    hooks:
+      - id: black
+        name: Run Black to format Python files
```
ScottTodd, line 15: There are some other hooks that we'll want but this is good as a starting point.

* https://github.com/iree-org/iree-test-suites/tree/main/linalg_ops has C++ source code that will be formatted with clang-format (https://github.com/iree-org/iree/blob/e8ee4801eda1babed0c513e941b88df3c3e25689/.pre-commit-config.yaml#L33-L41)
* Trailing whitespace, end of file fixer, and forbid tabs are generally useful
* check-yaml is useful for workflow files
* mypy type checking will be useful for python, but we can also have that run as part of the tests instead of lint

ScottTodd (approved):

PR #83, opened 2025-02-27
Fixing Litert Models test to reflect import error

geomin12:
Since tensorflow has released a new version, the tests are currently failing due to an import error. Updating the tests to reflect expected import error and xfail on those tests

### DIFF
--- a/litert_models/requirements.txt
+++ b/litert_models/requirements.txt
@@ -11,4 +11,4 @@ pytest-xdist
 # ai-edge-litert
 # tflite-runtime
 kagglehub
-tensorflow
+tensorflow<=2.18.0
--- a/litert_models/requirements.txt
+++ b/litert_models/requirements.txt
@@ -11,4 +11,4 @@ pytest-xdist
 # ai-edge-litert
 # tflite-runtime
 kagglehub
-tensorflow<=2.18.0
+tensorflow
--- a/litert_models/tests/kaggle/tensorflow/mobilenet_v1_test.py
+++ b/litert_models/tests/kaggle/tensorflow/mobilenet_v1_test.py
@@ -12,12 +12,12 @@ from ....utils import *
 
 
 # https://www.kaggle.com/models/tensorflow/mobilenet-v1/tfLite/0-25-224
-@pytest.mark.xfail(raises=IreeCompileException)
+@pytest.mark.xfail(raises=IreeImportTfLiteException)
 def test_mobilenet_v1_0_25_224(tflite_import_and_iree_compile):
     tflite_import_and_iree_compile("tensorflow/mobilenet-v1/tfLite/0-25-224")
 
 
 # https://www.kaggle.com/models/tensorflow/mobilenet-v1/tfLite/0-25-224-quantized/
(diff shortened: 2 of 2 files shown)


litert_models/requirements.txt
```diff
@@ -11,4 +11,4 @@ pytest-xdist
 # ai-edge-litert
 # tflite-runtime
 kagglehub
-tensorflow
+tensorflow<=2.18.0
```
ScottTodd, line 14: > to make tests pass

The tests are all failing right now and will continue to fail for at least another month until a new import API is added and we migrate to it as part of https://github.com/iree-org/iree/issues/19777. They are just marked as XFAIL for an outdated reason: https://github.com/iree-org/iree-test-suites/blob/f88f598252255147fcc76d092b1478572fbb787b/litert_models/tests/kaggle/tensorflow/mobilenet_v1_test.py#L14-L23

Per my comment on the tracking issue:

> We can add new error handling to iree-import-tflite (for as long we keep it around) and change the test suite to treat failures there as IreeImportTfLiteException or a new exception type.

The way I imagined this we would continue to test the latest versions of IREE and TensorFlow packages and update the test suites to expect the latest failures (failed to import, not failed to compile). This approach of pinning to an older version of the TensorFlow package (which we recommend at https://iree.dev/guides/ml-frameworks/tflite/) does retain the "import succeeded, compile failed" error status, which is reasonable for now.

I'd suggest amending the PR title and description. The fact that a prerelease version is installed is mostly irrelevant - that version is coming to stable soon and it will break in exactly the same way.

ScottTodd: I've also been ignoring the failures because TFLite support _is_ broken right now, so having the test suite reflect that could help set community expectations better.

litert_models/tests/kaggle/tensorflow/mobilenet_v1_test.py
```diff
@@ -12,12 +12,12 @@
 
 
 # https://www.kaggle.com/models/tensorflow/mobilenet-v1/tfLite/0-25-224
-@pytest.mark.xfail(raises=IreeCompileException)
+@pytest.mark.xfail(raises=IreeImportTfLiteException)
```
ScottTodd, line 15: Oh, nice that this actually worked. I was expected we'd need to change https://github.com/iree-org/iree/blob/034bc30519ebe391f3fe2166ff2b148bb8721c12/integrations/tensorflow/python_projects/iree_tflite/iree/tools/tflite/scripts/iree_import_tflite/__main__.py#L14 too.

Ah, since we invoke the tool in a subprocess the NameError there gets rolled into a failure error code here: https://github.com/iree-org/iree-test-suites/blob/f88f598252255147fcc76d092b1478572fbb787b/litert_models/utils.py#L54-L73

If we called the Python API in-process then we'd need different error handling.

ScottTodd (approved):

ScottTodd:
LGTM once the PR title + description match what has changed.

ScottTodd:
Thanks for your patience with this! Welcome to the world of long support horizons and cross-company ecosystems 🫠

PR #84, opened 2025-02-27
Better subprocess logs and extending timeout

geomin12:
Prints more verbose logs (screenshot of the start snippet of logs, of course the logs are quite long):

Screenshot 2025-02-27 122539

Also extending timeout for composite action as it now covers both SD3 and SDXL tests


sharktank_models/ireers_tools/fixtures.py
ScottTodd, line 1: Do you want to backport these changes to https://github.com/iree-org/iree/blob/main/experimental/regression_suite/ireers_tools/fixtures.py or just go ahead and complete the migration from that code to this code?

The largest obstacle to migrating right now in my mind is how tests are configured. The IREE repo has these spec files for tests/benchmarks: https://github.com/iree-org/iree/tree/main/build_tools/pkgci/external_test_suite while this repository currently uses https://github.com/iree-org/iree-test-suites/tree/main/sharktank_models/quality_tests/external_test_files here: https://github.com/iree-org/iree-test-suites/blob/f88f598252255147fcc76d092b1478572fbb787b/sharktank_models/quality_tests/sdxl/unet_fp16_rocm.json#L61-L63 

If the files are really still needed (https://github.com/iree-org/iree/issues/20044) then developers will need to be able to edit them in the IREE repo and have the tests use them, without needing to make cross-repository changes.

geomin12: After some looking, it seems like these `.mlir` files are only used for these tests. Tbh, i think we can delete them from iree and have them officially migrated to this repo. When these files are added as default to iree, these files + configs can be removed.

for `fixtures.py`, i do see that the `test/pregenerated/test_llama2.py` and `test/pregenerated/test_ukernel.py` are using `iree-compile`, but those tests are not included in the pkgci runs and not being tested anywhere. 

My thought: removing .mlir tuner files for regression tests + `experimental/regression_suite` directory from iree repo, and doing the full migration to `iree-test-suite`. Once those tuner files are default in iree compiler, we can clean those up

ScottTodd: The files sometimes need edits as part of changing the compiler, like on https://github.com/iree-org/iree/commit/41115bba05960e563791ce6ed1af26093f4fab1e and https://github.com/iree-org/iree/commit/a1664e30a6a53850f689f32086a8b0c45bed327b. Splitting the files across repositories will make such changes harder to perform.

_I_ certainly want to improve the situation there so the files aren't needed, but that's what the compiler team has been living with for a while now. We can keep bending the infra to support that usage... or notify them of the upcoming changes and get back to a cleaner state.

geomin12: Ah I see! Okay, mapping the mlir files to the iree repo is possible with GH actions and I can add notes to let users know that is how those files are retrieved. Hopefully those tuner files will be default soon so we don't have to deal with this!

For the `fixtures.py` + `regression_suite` repo, I'll removed that entirely unless you want to keep those pregenerated tests

ScottTodd: > For the `fixtures.py` + `regression_suite` repo, I'll removed that entirely unless you see value in keeping those pregenerated tests. Although, I can simply also just migrated those pregenerated tests over to keep even though they aren't being used

Deleting those is fine with me.

sharktank_models/benchmarks/sdxl/e2e_rocm.json
```diff
@@ -18,7 +18,7 @@
         "--iree-llvmgpu-enable-prefetch",
         "--iree-hal-target-backends=rocm"
     ],
-    "mlir_file_path": "external_test_files/sdxl_pipeline_bench_f16.mlir",
+    "mlir_file_url": "https://raw.githubusercontent.com/iree-org/iree/refs/heads/main/build_tools/pkgci/external_test_suite/sdxl_pipeline_bench_f16.mlir",
```
ScottTodd, line 21: > And also! Pulling MLIR files from iree repo to be up to date

This won't work. If someone is changing the file on a branch and sends a pull request, the tests should use that file. See how the ONNX tests pass files to the pytest command: https://github.com/iree-org/iree/blob/064ed35d85e9348f9451acb3fe6e0798f6ae4c97/.github/workflows/pkgci_test_onnx.yml#L102-L108

The architecture should be:

* Models can be sourced from a variety of locations, ideally provided in stable formats (e.g. ONNX, TFLite, GGUF files on HuggingFace or Kaggle)
* In this repository, test suites define test cases for models _given a set of configurations_
* Other repositories, like https://github.com/iree-org/iree, provide configurations to the test suite, choose which subsets of the test cases they want to run, and decide how they want to handle the outcomes (e.g. strict XFAIL/XPASS to be notified when a test is newly failing or newly passing)

Settings used to compile, including spec files like this, are a part of the configuration. The test suite repository can provide default configurations for utility, or source some configurations from the IREE repo like this, but external users like the IREE repository need to be able to provide the full configuration themselves.

---

Can you move these changes to a new PR so we can keep the discussion focused on the logging changes and land those first? Rule of thumb (which I sometime break, but do so intentionally): if a PR title has "and" in it, split it up.

ScottTodd (changes requested):

sharktank_models/benchmarks/sdxl/e2e_rocm.json
geomin12: ah i see! makes sense to allow custom config files. it's all making sense! okay cool, let me split this up then get that update rolling

ScottTodd (approved):

PR #89, opened 2025-03-19
Batched llama31 tests

renxida:
Include a test with batched llama 31 compile + numerics; also includes scripts to set up a venv and re-generate mlir files.


renxida:
Main questions:
- we good to just adjust the tolerance for the xfailed tests? Maybe to e-1?
- how do the asset-generation scripts look?

rsuderman (approved):

PR #92, opened 2025-03-27
Fixing sharktank_models test suite bugs.

geomin12:
- Fixing path for `job_summary.json`, artifacts and tests
- Fixing llama, clip, and llama_batched LFS path, as well as fixing llama_batched new flags
- Fixing Git LFS (composite actions is not a fan of pulling LFS files)
- Adding new flags from iree repo
- When an empty test suite runs, tests fail. Fixing that so it passes (happens with CPU benchmark tests)


sharktank_models/action.yml
```diff
@@ -19,21 +19,39 @@ inputs:
...
+    description: "Directory of the iree test files"
 
 runs:
   using: "composite"
   steps:
-    - name: Checkout repository 
+    - name: Checkout IREE repository
       uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2 
       with:
         clean: false # doesn't clean up the venv created before this composite action
```
ScottTodd, line 39: This is error prone - always assuming that this action will be run from the IREE repository.

What does this action actually need from that repository? Can those files be passed in to this workflow directly instead?

sharktank_models/benchmarks/conftest.py
```diff
@@ -91,6 +97,10 @@ def pytest_sessionfinish(session, exitstatus):
                 )
                 print("\n" + table_data, file=job_summary)
 
+    # on default, if no tests are collected, it exits with error. Instead, return with success status
+    if exitstatus == NO_TESTS_COLLECTED:
+        session.exitstatus = TEST_SUCCESS
```
ScottTodd, line 102: Why? We should avoid having silent failures, and running zero tests (e.g. on a CI workflow) is a failure that we shouldn't ignore.

We go out of our way to avoid the silent failure mode when using ctest: https://github.com/iree-org/iree/blob/5c7c1f2a3c1d3cab7e6c8f659a33fbf164fd929d/build_tools/cmake/ctest_all.sh#L157.

sharktank_models/benchmarks/conftest.py
```diff
@@ -15,6 +15,10 @@
 
 THIS_DIR = Path(__file__).parent
 sku = os.getenv("SKU", default="mi300")
+github_action_path = os.getenv("GITHUB_ACTION_PATH", str(THIS_DIR))
```
ScottTodd, line 18: Pass file paths in through some configuration or abstract this behind a github-actions utility script. This part of the test runner should not care what particular CI provider (if any) it is running on.

sharktank_models/benchmarks/conftest.py
```diff
@@ -15,6 +15,10 @@
...
+backend = os.getenv("BACKEND", default="rocm")
```
ScottTodd, line 19: Can we default to CPU instead?

sharktank_models/action.yml
```diff
@@ -19,21 +19,39 @@ inputs:
...
+  BACKEND:
+    description: "Type of backend to test"
   EXTERNAL_TEST_FILE_DIRECTORY:
     description: "The directory of test files and configurations"
   QUALITY_TEST_DIRECTORY:
     description: "The directory of quality tests"
   BENCHMARK_TEST_DIRECTORY:
     description: "The directory of benchmark tests"
+  IREE_TEST_FILES:
+    description: "Directory of the iree test files"
```
ScottTodd, line 31: There are enough variables in this file now that they should have documentation. Can you add a comment to the top of this file, or inline, explaining what is expected from each variable?

For example, provide sample values for `TARGET`, `GPU`, and `BACKEND` (these are overlapping, so be explicit here to show how they complement one another). Then for these directory variables, explain what files are expected to exist at each path, if the paths can be omitted to skip those types of tests, etc.

We may also want to instead provide all of these through a configuration file instead of individual variables, as is already done for the other test suites.

sharktank_models/llama3.1/test_llama.py
```diff
@@ -15,15 +15,16 @@
 block_size = 32
 page_size = kv_size * block_size
 
-THIS_DIR = pathlib.Path(__file__).parent
-llama_mlir = str(THIS_DIR / "assets/toy_llama.mlir")
-llama_irpa = [str(THIS_DIR / "assets/toy_llama.irpa")]
+THIS_DIR = pathlib.Path(__file__).parent.parent
+ASSET_PATH = os.getenv("ASSET_PATH", default=str(THIS_DIR)) + "/llama3.1"
+llama_mlir = f"{ASSET_PATH}/assets/toy_llama.mlir"
+llama_irpa = [f"{ASSET_PATH}/assets/toy_llama.irpa"]
```
ScottTodd, line 21: nits:

* Could wrap `ASSET_PATH` in `Path()` to avoid having a loose string object mixed with Pathlib objects
* Why is mlir a single file put irpa a list? Should irpa be plural?

ScottTodd (changes requested):

sharktank_models/llama3.1/test_llama.py
geomin12: seems like it's in an array for the below pytest.fixture to match other array arguments! and sounds good, made the Path change!

ScottTodd (changes requested):

sharktank_models/action.yml
```diff
@@ -8,32 +8,49 @@ name: IREE Sharktank Models Tests
...
-    - name: Checkout repository 
+    # Pulling IREE Test Suites specifically for Git LFS files
+    - name: Checkout IREE Test Suites repository
       uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2 
       with:
-        clean: false # doesn't clean up the venv created before this composite action
-    
+        repository: iree-org/iree-test-suites
+        path: iree-test-suites
+        lfs: true
```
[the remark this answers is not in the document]
geomin12: Yeah, so it currently uses the default branch (although now that I think about it, it should always use that corresponding hash, so let me fix that)

This feature may not exist yet (I searched everywhere for any docs about this), but composite actions doesn't support LFS yet so this is the workaround :/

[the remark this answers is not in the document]
ScottTodd: It feels like you're working against the grain here. Is a composite action actually viable? I don't mind duplicating 100 lines of yaml if we get the right behavior and actions are easy to maintain.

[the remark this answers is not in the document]
geomin12: yeahhhh, the only con is the LFS problem! It does make it much cleaner.

However IREE currently just clones iree-test-suite and runs the tests (LFS included), so it's practically the same except JSON file configuration. My thought is we can either keep this format until LFS support comes for composite action, or get rid of composite action and just do the good ole clone and run.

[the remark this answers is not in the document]
ScottTodd: Do whatever is simplest and works first (no composite action?), then refactor once fully migrated. I don't want us maintaining two versions of the same tests for any longer than we need to.

geomin12:
feel free to review other things, looking at the error now. looks like for CPU it doesn't collect any tests and fails

ScottTodd (approved):

PR #94, opened 2025-04-02
Llama 8b f16 regression test using random weights and inputs

aviator19941:
This PR adds regression tests for llama 8b f16 prefill and decode using randomized llama 8b f16 weights and inputs. We use random llama 8b f16 weights in order to download a public version of the weights (the original weights are gated and require access from the model authors from the huggingface repo).


aviator19941:
@geomin12 It seems like the new json files didn't run, am I missing something?

PR #112, opened 2025-09-23
Added check such that if -k option is provided then it will download …

pdhirajkumarprasad:
Added check such that if -k option is provided then it will download artifacts for those models only


pdhirajkumarprasad:
Also @Groverkss is working on upgrading the pytest framework so has asked me not to push these changes, his changes will land in few days

PR #119, opened 2025-10-07
Golden time update for clip CPU

newling:
First change was https://github.com/iree-org/iree-test-suites/commit/3b0628c55b059361a9b9dd85c133c74cf0ae9416
Second change was https://github.com/iree-org/iree-test-suites/commit/a8e9673a60ca9f324e030c980d41194a61bf3885

Current failure I see is 

INFO     model_benchmark_run:model_benchmark_run.py:226 sdxl clip benchmark time: 38922.793748500015 ms (golden time 2850.0 ms)

See https://github.com/iree-org/iree-test-suites/actions/runs/18317198606/job/52160587711

Searching for '2850' (the golden time the test is still failing) with I found it in a different file. Did the first 2 changes change the wrong value? Bumping this golden value and returning the pytorch one to its original value.

### DIFF
--- a/sharktank_models/benchmarks/sdxl/clip_cpu.json
+++ b/sharktank_models/benchmarks/sdxl/clip_cpu.json
@@ -16,7 +16,7 @@
         "cpu": 1.1
     },
     "golden_time_ms": {
-        "cpu": 2850.0
+        "cpu": 40000.0
     },
     "golden_dispatch": {
         "cpu": 2100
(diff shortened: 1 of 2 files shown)


newling:
> EDIT: on second look, I think you are right. Thanks for pointing it out.

These golden values are very confusing (I have so far found 3 'clip cpu' golden times across 2 repos!) I'll land this, give it a try, and if I still don't have the CI tests passing will dive deeper.

Issue #116, opened 2025-10-03
Investigate regression in clip_cpu


amd-eochoalo:
See [here for an example of a failing job](https://github.com/iree-org/iree-test-suites/actions/runs/18226081135/job/51897601776).

hanhanW:
Can you provide the link for the corresponding integrate PR? Otherwise, it is hard to track the changes.

```
AssertionError: FAILURE: check 32374.841299000003 <= 3135.0000000000005: sdxl clip benchmark time should not regress more than a factor of 1.1
```

The regression is big enough that you can bisect it easily on most CPUs, I think.

PR #123, opened 2025-10-14
Update CPU flags to follow https://github.com/iree-org/iree setup

hanhanW:
The regression was caused because of the data-tiling flip. It's disabled by default, so we have to explicitly add the flag.

The revision also refreshes the CPU flags to follow what we've been using in iree repo; update the golden values for torch models.

The sharktank golden values are back; additionally, the revision tightens the golden values a bit.

Fixes https://github.com/iree-org/iree-test-suites/issues/116

### DIFF
--- a/sharktank_models/benchmarks/sdxl/clip_cpu.json
+++ b/sharktank_models/benchmarks/sdxl/clip_cpu.json
@@ -16,10 +16,10 @@
         "cpu": 1.1
     },
     "golden_time_ms": {
-        "cpu": 40000.0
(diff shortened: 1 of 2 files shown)


hanhanW:
Okay, it fixes the issue. I'm updating the golden values

amd-eochoalo (approved):
The golden time for clip_cpu is now les than what it was before. So happy to approve. I can't comment on adding / removing some of the flags so you may want to wait for other reviewers, but I think it makes sense.

PR #124, opened 2025-10-20
Allow using MLIR files checked into the repo (using git-lfs)

MaheshRavishankar:
This change is to allow using MLIR files that are checked into the repository using GIT LFS. To support this adding a new "type" field to the compilation modules JSON that says how to get the MLIR.

### DIFF
--- a/torch_models/examples/modules/scheduled_unet_cpu.json
+++ b/torch_models/examples/modules/scheduled_unet_cpu.json
@@ -1,4 +1,5 @@
 {
+  "type": "azure",
   "mlir": "https://sharkpublic.blob.core.windows.net/sharkpublic/sai/sdxl-scheduled-unet/model.mlir",
   "compiler_flags": [
       "--iree-hal-target-device=local",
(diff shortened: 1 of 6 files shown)


torch_models/pytest_iree/git_lfs.py
```diff
@@ -0,0 +1,34 @@
...
+import logging
+import subprocess
+from pytest_iree.artifact import Artifact
+
+logger = logging.getLogger(__name__)
+
+class GitLfsArtifact(Artifact):
+    """
+    Represents an artifact that is stored in the repo using git lfs.
+    """
```
Groverkss: Can you add details on the directory structure it creates like other artifacts have?

torch_models/examples/modules/scheduler_cpu.json
```diff
@@ -1,10 +1,9 @@
 {
+  "type": "azure",
   "mlir": "https://sharkpublic.blob.core.windows.net/sharkpublic/sai/sdxl-scheduled-unet/sdxl_unet_pipeline_bench_f16.mlir",
```
Groverkss: I think we can add this scheduler in lfs and use it directly as a test.

torch_models/examples/modules/scheduled_unet_cpu.json
```diff
@@ -1,11 +1,11 @@
 {
+  "type": "azure",
```
Groverkss: Can you add "type" to README.md and explain how it works and expects?

Groverkss (commented):
Mostly LGTM, would be good to land this

PR #127, opened 2025-11-06
[torch_models] Use min time instead of mean time

amd-eochoalo:
The commit-id referenced in IREE pointed to this PR https://github.com/iree-org/iree/blob/main/.github/workflows/pkgci_test_torch.yml#L72 . But it should be merged to main.


Groverkss (approved):
IREE CI is already using this commit, ok to land without much review

PR #134, opened 2025-11-19
Remove QOL flags

IanWood1:
Removes the three QOL flags. These flags are overly verbose and don't appear to offer much benefit to developers.

See this workflow run with 100k lines of debug output: https://github.com/iree-org/iree/actions/runs/19488963275/job/55777673696?pr=22341

### DIFF
--- a/sharktank_models/benchmarks/sdxl/clip_cpu.json
+++ b/sharktank_models/benchmarks/sdxl/clip_cpu.json
@@ -19,7 +19,7 @@
         "cpu": 2755.0
     },
     "golden_dispatch": {
-        "cpu": 2089
+        "cpu": 2130
     },
     "golden_size": {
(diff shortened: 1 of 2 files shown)


IanWood1:
> Ok, something is off.
> 
>     1. I think the dispatch count should not have been updated with the QOL flag drops. When I approved, that wasnt there.
> 
>     2. The dispatch count should not depend on machine at all. Something really bad is happening here.

Sorry for updating them post approval. From what I could tell, CI isn't using a pinned version of IREE. I think a newer version was used in CI, breaking the dispatch counts.

Looking at https://github.com/iree-org/iree-test-suites/actions/runs/19511544463/job/55852291282?pr=136, the failure is due to time and not dispatch counts.

PR #138, opened 2025-11-30
[torch_ops] Increase acceptable tolerance for select tests 

efric:
The tolerances for these tests are unnecessarily tight given the current method with `np.allclose`. Tests that include the addition of a bias with inputs in range [-1, 1] are particularly sensitive. The addition doesn't contribute much additional error but because it changes the output values, it affects np.allclose's effective tolerance which is measured by:

`|a - b| ≤ atol + rtol × |b|`

Thus for large values rtol dominates whereas for small values atol dominates.

For `A @ B + C`, with inputs in `[-1, 1)` as is currently, C can cancel the result of the matmul at certain indices which results in outputs very close to 0 and thus the atol term dominates when measuring tolerance. However, `atol=1e-8` is too tight. In a [case study based on a CI failure](https://github.com/iree-org/iree/actions/runs/19586592340/job/56208325511), the error at the worst failing index is `1.33e-06`, which is well within its acceptable bound* of `1.04e-04` but fails due to tight atol.

For convenience we can keep using `np.allclose` but use `atol=1e-5` since that's closer to the scale of actual FP errors for this problem. It covers acceptable observed errors with margin to spare.

*Based on Higham's forward error bound for matrix multiplication (Accuracy and Stability of Numerical Algorithms):
Given:

`|Ĉ_ref - C_exact|\_{i,j} ≤ (γ_n|A||B|)\_{i,j} where γ_n = nu/(1 - nu)`, `C_ref` is the PyTorch CPU reference, and `u = np.finfo(np.float32).eps / 2`.

Then for an implementation, it follows that:

`|Ĉ_ref - Ĉ_impl| ≤ |Ĉ_ref - C_exact| + |C_exact - Ĉ_impl| ≤ 2γ_n|A||B|`


efric:
I'm also curious to learn a bit on what the existing input ranges/tolernaces for each respective test are based on, since they vary depending on the test

kuhar:
@bjacob could you check if the math makes sense?

bjacob:
> I'm also curious to learn a bit on what the existing input ranges/tolernaces for each respective test are based on, since they vary depending on the test

The rationale is that we haven't invested in any such analysis here. We are just trying to catch any regression or unintended change to numerics. It's OK if an overly strict test needs to be relaxed. So we just default to as tight tolerances as we get away with, and relax as we go. So this PR is the system working as intended. Thanks for the careful analysis!

Can you share some link or detail about the test failure that motivated this change?  Ah I see it's what the "case study" links to - it wasn't clear to me that that term would mean a CI failure. Maybe clarify.

amd-eochoalo:
> I'm also curious to learn a bit on what the existing input ranges/tolerances for each respective test are based on, since they vary depending on the test

> The rationale is that we haven't invested in any such analysis here. We are just trying to catch any regression or unintended change to numerics.

I didn't spend too much time looking into the required tolerances for all tests. The only tests which had a tolerance explicitly set are those for which the default tolerance was not met. I then followed a similar analysis (looked at machine epsilon and number of operations) to find a reasonable error.

> So we just default to as tight tolerances as we get away with, and relax as we go. 

This is what I thought at the time but did not explicitly state.

PR #139, opened 2025-11-30
[torch_ops] Fix erroneous types for GeluABPlusC f16 test

efric:
This seems to be using the wrong types.


kuhar (approved):

amd-eochoalo:
Let me run it on CI, that way the result will be ready earlier today :)

PR #142, opened 2025-12-11
Changed onnx_model repo to hf

prosenjitdhole:
Changed onnx model repo path to hf.
Ref : https://github.com/iree-org/iree-test-suites/issues/101

Please note the assumptiosns : 
- Only legacy models were used, 
- Currently it does not have an option to fall back to old git repo. 

ci-extra: test_torch

PR #145, opened 2026-01-12
Small fixes to the CI

amd-eochoalo:
* Couple of small fixes to torch_ops.
* Replace ONNX models configurations file with those in IREE.

ONNX models is turned off in iree-test-suite's CI, however today we needed to run the test suite to test changes in how the model data is downloaded. These changes cannot be tested easily in iree-test-suite because they have been turned off, have been failing for a long time. Since ONNX models are passing in IREE and the only target that will be running in iree-test-suite's CI is the CPU backend, then I think it can be a good idea to copy these here.

This would have allowed one to easily test the changes in #142 here without needing to create a branch in IREE to point to #142 commit ref for testing the CPU target.

### DIFF
--- a/onnx_models/README.md
+++ b/onnx_models/README.md
@@ -1,7 +1,7 @@
 # ONNX Model Tests
 
 This test suite exercises ONNX (Open Neural Network Exchange: https://onnx.ai/)
-models. Most pretrained models are sourced from https://github.com/onnx/models.
+models. Most pretrained models are sourced from https://huggingface.co/onnxmodelzoo/legacy_models/.
 
 Testing follows several stages:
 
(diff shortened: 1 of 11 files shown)


amd-eochoalo:
@Groverkss THIS_DIR was not define in that file https://github.com/iree-org/iree-test-suites/pull/145/changes/77d4427b7e24699ac7c586669f757f2a2f5b67c8

and the correct path for the mlir file was composed by an undefined variable https://github.com/iree-org/iree-test-suites/pull/145/changes/db72c1eabae9133de3ed2fe8a7b9ee80bfa21d9f . so it got corrected.

PR #159, opened 2026-02-09
[torch_ops] Use rocprofv3 when available and fix benchmark code

### LINKED ISSUE #131: [torch_ops] Benchmarks
We would like to have some microbenchmarks defined at the same abstraction level as torch operations. The infrastructure behind these microbenchmarks should follow a similar design (and reuse code wherever possible) as the torch operation correctness tests. Other considerations have been collected somewhat loosely here:

* `iree-run-module`/`iree-benchmark-module` may have limited applicability here since timing it will also include time spent on CPU and not on kernel dispatches.
* `iree-benchmark-executable` would be ideal to use, as it is target agnostic, but it currently lacks GPU timestamps.
* `torch.profiler` brings torch as a dependency and we would like these tests to not depend on `torch`.
* `rocprofv3` target specific.
* `tracy` would need a debug build of `iree-run-module`.

At the moment, the work being done focuses on using `rocprofv3` and will generalize to use `iree-benchmark-module`/`iree-benchmark-executable` for other targets.

amd-eochoalo:
1. This change will use rocprofv3 to calculate the minimum time spent on the kernel dispatch (assumes only one kernel dispatch that may be called multiple times per test). This is similar to iree-benchmark-module which will report the average time for a single kernel dispatch. However, unlike iree-benchmark-module, which measures the wall clock-time per iteration including VM overhead, rocprofv3 will exclude host overhead and report GPU execution time only. The minimum is selected instead of the mean of all dispatches which is a better indication of steady state execution. rocprofv3 will only be used when  "--iree-hal-target-device=hip" is one of the compilation flags. Please note that we still use iree-benchmark-module to actually execute the kernel, but this should be fine since it allows us to reuse iree-benchmark-module's capacity to run many times to provide some statistics. (Unlike running `rocprofv3 -- iree-run-module` which would provide a single measurement)
2. This PR also fixes a couple of errors. renames `run_cmd` -> `cmd` and `test_file` -> `test_dir`. 

The installation of the rocm package is not shown here. This is because in iree-test-suite the runs only execute in CPU. On iree's CI, the rocm package installation should be part of the yaml file as it depends on the GPU. See here for a successful run on IREE's CI targeting gfx1201 https://github.com/iree-org/iree/actions/runs/21836654936/job/63011830015

Closes #131

[gfx942](https://github.com/iree-org/iree/actions/runs/21840139718/job/63022782616) on CI consistently fails to profile the program. The data always comes up empty. For this reason the option is only enabled as a command line option to pytest that IREE's CI must specify.

### DIFF
--- a/.github/workflows/test_torch_ops.yml
+++ b/.github/workflows/test_torch_ops.yml
@@ -74,9 +74,7 @@ jobs:
             --self-contained-html \
             --report-log=/tmp/torch_ops_cpu_logs.json \
             --config-files=${CONFIG_FILE_PATH} \
-            --artifact-directory=${CACHE_DIR}/torch_ops_artifacs \
-            --benchmark-with-rocprofv3
-
+            --artifact-directory=${CACHE_DIR}/torch_ops_artifacs
       - name: Upload HTML report
         uses: actions/upload-artifact@b7c566a772e6b6bfb58ed0dc250532a479d7789f # v6.0.0
         with:


torch_ops/common.py
```diff
@@ -338,8 +338,47 @@ def iree_run_module(self, iree_run_flags):
...
+        cmd = [
+            "rocprofv3",
+            "--kernel-trace",
+            "--stats",
+            "--output-file",
+            rocprofv3_output,
+            "--output-format",
+            "json",
+            "--",
+            "iree-benchmark-module",
```
kuhar: Do we need to run iree-benchmark-module under rocprof? The alternative I'm thinking about it so invoke iree-run-module (either via cli or using python bindings)

amd-eochoalo: I'm open to change it, but I think that iree-benchmark-module already provides a nice way to say "run until you achieve some statistical certainty about the measurements collected" (which I understand happens automatically with iree-benchmark-module). I think running with the python bindings may be a bit more cumbersome because rocprof will need profile the parent process and there might be a race condition depending on how much we wait for the child process to start and when we dispatch kernels to the GPU via the bindings.

kuhar (approved):

PR #162, opened 2026-02-17
[torch_models] Move pytest_iree out of torch_models.

amd-eochoalo:
When moving torch_ops and onnx_ops to azure, we will need access to some of the files and classes defined in pytest_iree currently only residing under torch_models. This will prevent duplication of code.


amd-eochoalo:
Successful run in IREE : https://github.com/iree-org/iree/actions/runs/22113738721/job/63919566473

kuhar (approved):

PR #174, opened 2026-04-22
Remove deleted flag and update golden binary size

amd-eochoalo:
* "--iree-global-opt-experimental-disable-conv-generalization" was deleted in https://github.com/iree-org/iree/pull/23377
* updates golden binary size


MaheshRavishankar (approved):
Hmmm... binary size went up. That usually means that it got slower as well? Do you see that?

amd-eochoalo:
> Hmmm... binary size went up. That usually means that it got slower as well? Do you see that?

I'll look into it!

amd-eochoalo:
@MaheshRavishankar , looking into it, these CPU tests are running on github runners and to avoid a huge variance / noise we are not tracking the golden time in iree-test-suite.

PR #177, opened 2026-06-18
Add CPU quality regression tests (ResNet50, Qwen3-0.6B, Whisper-tiny.en)

roberto-laudani:
Adds an end-to-end CPU quality regression for `microsoft/resnet-50`, `Qwen/Qwen3-0.6B`, and `openai/whisper-tiny.en`. IREE compiles the committed `model.mlir`, fetches the externalized parameters from the Hugging Face Hub (pinned by revision), runs on `llvm-cpu`, and compares the output logits against the committed reference.


roberto-laudani:
The red "Test Torch Models" check is the pre-existing `scheduled_unet_compstat_cpu` binary-size failure, tracked in #178 and is unrelated to this PR. The resnet50 test added here passes.

torch_models/qwen3-600m/generate.py
```diff
@@ -0,0 +1,109 @@
...
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        "--out-dir",
+        type=Path,
+        default=Path(__file__).parent / "artifacts",
+    )
+    args = parser.parse_args()
+    args.out_dir.mkdir(parents=True, exist_ok=True)
+
+    print(f"Loading model {MODEL_ID}@{MODEL_REVISION} (f32)")
```
chrsmcgrr:
```suggestion
    print(f"Loading model {MODEL_ID}@{MODEL_REVISION} (f32)", file=sys.stderr)
```
For scripts and binaries if its not an output and just logging info use stderr. Stdout is reserved for tool output.

chrsmcgrr (commented):
Looks good! Would it also be possible to have the binary files on HF? We would want to avoid checking in binary files because they can't be diffed. Changing it later will cause a new copy in the history.

chrsmcgrr:
@schuermans-roofline will help with a review :)

pytest_iree/module.py
```diff
@@ -86,12 +86,22 @@ def get_compstats(self) -> dict:
...
         module_data = json.loads(self.module_json.read_text())
-        mlir_url = module_data["mlir"]
-        assert (
-            "blob.core.windows.net" in mlir_url
-        ), "Only Azure Blob Storage is supported currently."
-        self.mlir_artifact = AzureArtifact(self.artifact_base_dir, mlir_url)
-        self.mlir_artifact.join()
-        return self.mlir_artifact.path
+        mlir_ref = module_data["mlir"]
+        if "blob.core.windows.net" in mlir_ref:
```
schuermans-roofline: I think this check might be a bit too unspecific. (For example, if somebody downloaded some files from there and named the local folder like the host name.) Would it work to check for `.startswith("https://")` in addition? Or even a regexp-based check?

pytest_iree/huggingface.py
```diff
@@ -0,0 +1,48 @@
...
+            repo_type=self.repo_type,
+        )
+        self._path = Path(local_path)
+        logger.info(
+            f"  Using HuggingFace artifact '{self.repo_id}/{self.filename}'"
+            f"{f'@{self.revision}' if self.revision else ''} at '{self._path}'"
+        )
+
+    def __str__(self):
+        return f"hf://{self.repo_id}/{self.filename}"
```
schuermans-roofline: Is this string form parsed anywhere again or is if for information only and can be changed without much hassle? If the latter, then I'd suggest to add more information to it.
```suggestion
        at_rev = "" if self.revison is None else "@" + self.revision
        return f"hf://{sef.repo_type}/{self.repo_id}{at_rev}/{self.filename}"
```

pytest_iree/test_base.py
```diff
@@ -112,9 +131,22 @@ def _get_arg_strings(self, json_field: str) -> list[str]:
...
+                )
+                artifact.join()
+                strings.append(f"@{str(artifact.path.absolute())}")
+            elif file is not None:
+                file_path = Path(file)
+                if not file_path.is_absolute():
+                    file_path = self.external_file_directory / file_path
+                strings.append(f"@{str(file_path.resolve())}")
             assert (
                 len(strings) > 0
```
schuermans-roofline, line 148: Now that the string collection got more complicated, it might make sense to add `and len(strings) <= 2` here, so nothing is producing sth like `<value>=@<file>=@<huggingface>` if a future change of the if/elif part introduces a bug.

pytest_iree/module.py
schuermans-roofline: It might make sense to check a https URL first and only check in a second step for the hostname part being on blob.core.windows.net, so the error "only supports blob.core.windows.net" can be kept. I think the current code would check for a local path if it finds `https://some.other.host/some_file`

schuermans-roofline (changes requested):
Looks basically good to me. I just had a few tiny remarks, I would like you to consider.

pytest_iree/huggingface.py
roberto-laudani: Yes, it is for information only.

schuermans-roofline (approved):
LGTM. Thanks.

